PR #2, opened 2019-12-17
Update Readme for jaeger chart

scottrigby:
See #1 for context.

This chart should initially fail CI, until I push another commit that bumps the version, then it should pass.

Signed-off-by: Scott Rigby <scott@r6by.com>

### DIFF
--- a/charts/jaeger/README.md
+++ b/charts/jaeger/README.md
@@ -68,10 +68,16 @@ data:
 
 ## Installing the Chart
 
+Add the Jaeger Tracing Helm repository:
+
+```console
+$ helm repo add jaegertracing https://jaegertracing.github.io/helm-charts
+```
+
 To install the chart with the release name `myrel`, run the following command:
 
 ```bash
-$ helm install incubator/jaeger --name myrel
+$ helm install jaegertracing/jaeger --name myrel
 ```
 
 After a few minutes, you should see a 3 node Cassandra instance, a Jaeger DaemonSet, a Jaeger Collector, and a Jaeger Query (UI) pod deployed into your Kubernetes cluster.
@@ -79,7 +85,7 @@ After a few minutes, you should see a 3 node Cassandra instance, a Jaeger Daemon
 IMPORTANT NOTE: For testing purposes, the footprint for Cassandra can be reduced significantly in the event resources become constrained (such as running on your local laptop or in a Vagrant environment). You can override the resources required run running this command:
 
 ```bash
-helm install incubator/jaeger --name myrel --set cassandra.config.max_heap_size=1024M --set cassandra.config.heap_new_size=256M --set cassandra.resources.requests.memory=2048Mi --set cassandra.resources.requests.cpu=0.4 --set cassandra.resources.limits.memory=2048Mi --set cassandra.resources.limits.cpu=0.4
+helm install jaegertracing/jaeger --name myrel --set cassandra.config.max_heap_size=1024M --set cassandra.config.heap_new_size=256M --set cassandra.resources.requests.memory=2048Mi --set cassandra.resources.requests.cpu=0.4 --set cassandra.resources.limits.memory=2048Mi --set cassandra.resources.limits.cpu=0.4
 ```
 
 > **Tip**: List all releases using `helm list`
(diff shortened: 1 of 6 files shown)


scottrigby:
As expected:

Screen Shot 2019-12-16 at 10 52 46 PM

Screen Shot 2019-12-16 at 10 53 08 PM

scottrigby:
I thought perhaps this was the same issue as https://github.com/helm/chart-releaser-action/issues/6 but the fix for that didn't help this. We can follow-up with this in https://github.com/helm/chart-testing-action and fix as part of this PR.

charts/jaeger/README.md
```diff
@@ -68,18 +68,24 @@ data:
```
yurishkuro: at L3, please change http://jaeger.readthedocs.io/en/latest/ to https://www.jaegertacing.io

charts/jaeger/README.md
```diff
@@ -68,18 +68,24 @@ data:
 
 ## Installing the Chart
 
+Add the Jaeger Tracing Helm repository:
+
+```console
+$ helm repo add jaegertracing https://jaegertracing.github.io/helm-charts
```
yurishkuro: Visiting https://jaegertracing.github.io/helm-charts/ shows 404. Should there be some landing page, maybe pointing back to this repo at minimum?

naseemkullah:
This might not be related, but FWIW all e2e tests for incubator/jaeger are failing as well:

https://github.com/helm/charts/pull/19638
https://github.com/helm/charts/pull/19525

cpanato:
the failing tests in the helm/chart repo was an infra issue, that is already fixed

charts/jaeger/README.md
scottrigby: A Helm Charts repo just needs an index.yaml that point to valid chart resources, and many don't have landing pages if you visit them (for example, the helm/charts stable repo landing page is just XML https://kubernetes-charts.storage.googleapis.com/). But it might be nice to add a friendly page 😄  I've seen some good ones, and gh-pages hosting method makes this very easy too

scottrigby:
The tests are running now (I was tired when I "fixed" the dependency repos issue last night). 

However now the ct-install test is failing, but with specifics: https://github.com/jaegertracing/helm-charts/runs/353223447 Can someone more familiar with Jaeger take a look?

I'm seeing:
```
Error: release jaeger-m6f1m5xi5g failed: timed out waiting for the condition
```

and events:

>  Warning  Unhealthy  28s   kubelet, chart-testing-control-plane  Liveness probe failed: Get http://10.244.0.5:14271/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
  Warning  Unhealthy  27s   kubelet, chart-testing-control-plane  Readiness probe failed: Get http://10.244.0.5:14271/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)

naseemkullah:
I think it is timing out because Cassandra which is a dependency takes too long to fire up, sometimes the required schema job times out (I had a PR open in the other repo to increase the deadline timeout for the job to 300s) 

It might be preferable that default values use in memory storage.

naseemkullah:
@scottrigby  Could you try increasing the chart tester timeout to `300`?

And this may be outside the scope of this PR but also increasing https://github.com/jaegertracing/helm-charts/blob/ed8fbe09cecbab02ea3bd3ce02c25af1e295763f/charts/jaeger/values.yaml#L95 to 300 will avoid headaches. (I had a PR open for this in previous repo)

scottrigby:
@naseemkullah I pulled in your commit from https://github.com/helm/charts/pull/19275, and also increased ct (helm install) timeout, which worked. Thanks for the tip there, I'm not familiar enough with the Jaeger application or chart to have known what the issue was without a lot of digging 👍

scottrigby:
CI for dependencies, and special timeouts for this specific chart are now fixed. Fixing up one more requested readme change

charts/jaeger/README.md
scottrigby: updated 👍

charts/jaeger/README.md
scottrigby: Let's look at that in a separate PR. Here's an example that's kind of nice and also easy https://charts.fluxcd.io/

PR #8, opened 2019-12-18
Made it possible to overwrite --reporter.grpc.host-port flag

ssoerensen:
It was not possible to change the --reporter.grpc.host-port to a custom hostname in case
the agent should send to a different service name.


naseemkullah:
Thanks @ssoerensen !

Issue #6, opened 2019-12-17
Landing page


naseemkullah:
Let's look at that in a separate PR. Here's an example that's kind of nice and also easy https://charts.fluxcd.io/

_Originally posted by @scottrigby in https://github.com/jaegertracing/helm-charts/pull/2_

cpanato:
we setup this for Mattermost: https://helm.mattermost.com/ simple

cpanato:
i liked the fluxcd maybe will clone :)

naseemkullah:
Solve via #10

PR #10, opened 2019-12-20
Landing page

naseemkullah:
Heavily inspired by @stefanprodan's  work for the Flux charts landing page, here is a PR to do this for Jaeger as well.

Is there a way of previewing this @scottrigby ?

We would need `charts.jaegertracing.io` added as a CNAME before merging

Would solve #6 

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,0 +1,1 @@
+charts.jaegertracing.io
--- a/CNAME
+++ b/CNAME
@@ -1,1 +1,0 @@
-charts.jaegertracing.io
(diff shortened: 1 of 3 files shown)


CNAME
```diff
@@ -0,0 +1 @@
+charts.jaegertracing.io
```
scottrigby: We'd need the Jaeger team to create this. You probably have already but for ref see https://help.github.com/en/github/working-with-github-pages/configuring-a-custom-domain-for-your-github-pages-site

scottrigby:
Re previewing https://help.github.com/en/github/working-with-github-pages/testing-your-github-pages-site-locally-with-jekyll

Issue #4, opened 2019-12-17
[jaeger] Use elastic chart as elasticsearch dependency


naseemkullah:
The optional elasticsearch dependency chart should be https://github.com/elastic/helm-charts/tree/master/elasticsearch

scottrigby:
Do you want to explore this as an alternative dependency in a PR?

naseemkullah:
Yes, sounds good, I'd prefer to completely remove the current deprecated elasticsearch dependency and swap in the new and improved dependency instead of manage values for both. Is that ok provided our disclaimer?

scottrigby:
Whichever chart dependency you prefer 😄

naseemkullah:
Solved via #11

PR #11, opened 2019-12-20
[jaeger] Use official elastic chart as elasticsearch dependency

naseemkullah:
Would solve #4

P.S. I also updated the README's helm install commands slightly... and removed some verbosity.

Also the condition moves form `provisionDataStore.elasticsearch` to the more standard `elasticsearch.enabled` , thoughts on that? We could do something similar once we use a different cassandra chart to depend on (https://github.com/bitnami/charts/tree/master/bitnami/cassandra seems like the best choice atm, but let's wait to see if Cassandra incubator chart gets hosted elsewhere)


naseemkullah:
bump

PR #13, opened 2019-12-21
Add env COLLECTOR_ZIPKIN_HTTP_PORT when collector.service.zipkinPort present

tiggerlee2:
Collector will not listen Zipkin port without the environment variable.


naseemkullah:
Thanks for the PR @tiggerlee2 !
Could you please GPG sign the commit, resolve conflicts and finally I think it would make sense to comment out `collector.service.zipkinPort` in values.yaml (and adjust the default value in README to `nil`) if we will wrap the ZIPKIN stuff in if statements

tiggerlee2:
@naseemkullah fixed as you said~

PR #18, opened 2019-12-24
[jaeger] refactor service accounts

naseemkullah:
When service accounts were originally added, the helm documentation regarding best practices on sa creation for multiple components was wrong. It has since been corrected and this approach follows the correct method


naseemkullah:
bump

PR #19, opened 2019-12-25
[jaeger] Use label helpers

naseemkullah:


charts/jaeger/templates/_helpers.tpl
```diff
@@ -31,6 +31,26 @@ Create chart name and version as used by the chart label.
...
+{{/*
+Common labels
+*/}}
+{{- define "jaeger.labels" -}}
+helm.sh/chart: {{ include "jaeger.chart" . }}
+{{ include "jaeger.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
```
pavelnikolov: Do we need `app.kubernetes.io/managed-by` label?

naseemkullah: It's among the recommended labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

Furthermore, `helm create` includes this as per: https://github.com/helm/helm/blob/12e0ab7e44eb02738e1128ac01ee0843e47b4e79/pkg/chartutil/create.go#L374

PR #22, opened 2020-01-03
[stable/jaeger] Fix indentation for cronjob affinity, nodeselector, tolerations

naseemkullah:


naseemkullah:
This is a bug fix. Currently, if trying to set any of these values for the cronjob, the install will fail since the indentation is wrong.

PR #30, opened 2020-01-15
Added extraSecretMounts option for TLS support

arpitjindal97:


arpitjindal97:
@naseemkullah  Can u check pr build logs? I wasn't able to find any issue here.

charts/jaeger/values.yaml
```diff
@@ -167,11 +179,17 @@ collector:
...
-  extraConfigmapMounts: []
+  extraSecretMounts: []
   # - name: jaeger-tls
   #   mountPath: /tls
   #   subPath: ""
-  #   configMap: jaeger-tls
+  #   secretName: jaeger-tls
+  #   readOnly: true
+  extraConfigmapMounts: []
+  # - name: jaeger-config
```
naseemkullah: what's this commented out configmount example, when would we use this?

arpitjindal97: Collector and agent both has an option of passing config-file

https://www.jaegertracing.io/docs/1.16/cli/

naseemkullah:
Thanks @arpitjindal97

PR #31, opened 2020-01-16
Added flexibility to change keyspace to the users

arpitjindal97:
Here is the discussion thread where it was decided to add this functionality

https://github.com/jaegertracing/jaeger-operator/pull/837

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.16.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
-version: 0.18.4
+version: 0.18.5
 keywords:
   - jaeger
   - opentracing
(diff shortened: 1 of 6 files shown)


arpitjindal97:
Here I figured out some thing, cassandra schema job container creates default schema, `jaeger_v1_dc1` but collector and query's default image uses schema `jaeger_v1_test`. 

Cassandra schema job container needs a change.

naseemkullah:
> Here I figured out some thing, cassandra schema job container creates default schema, `jaeger_v1_dc1` but collector and query's default image uses schema `jaeger_v1_test`.

I just realized that this will be a breaking change for anyone using the chart. We should probably address this soon before users get a chance to upgrade.

> 
> Cassandra schema job container needs a change.
> 
> I have made PR to jaeger repo also, to submit this fix.
> 
> [jaegertracing/jaeger#2020](https://github.com/jaegertracing/jaeger/pull/2020)

PR #37, opened 2020-01-22
[jaeger] Add security context to deployment charts in jaeger

mehta-ankit:

When deploying this chart in our cluster we found that we had to add security context to the deployments for them to run as non-root user.
So i created this PR where securityContext is added to the template/charts for deployment and its default value is provided in `values.yaml` file. These can be overwritten if someone uses jaeger chart as a dependency in their own values.yaml file.

### DIFF
--- a/charts/jaeger/values.yaml
+++ b/charts/jaeger/values.yaml
@@ -100,6 +100,8 @@ elasticsearch: {}
 
 ingester:
   enabled: false
+  podSecurityContext: {}
+  securityContext: {}
   annotations: {}
   image: jaegertracing/jaeger-ingester
   pullPolicy: IfNotPresent
@@ -138,6 +140,8 @@ ingester:
   extraConfigmapMounts: []
 
 agent:
+  podSecurityContext: {}
+  securityContext: {}
   enabled: true
   annotations: {}
   image: jaegertracing/jaeger-agent
@@ -191,6 +195,8 @@ agent:
   dnsPolicy: ClusterFirst
 
 collector:
+  podSecurityContext: {}
+  securityContext: {}
   enabled: true
   annotations: {}
   image: jaegertracing/jaeger-collector
@@ -279,6 +285,8 @@ collector:
 
 query:
   enabled: true
+  podSecurityContext: {}
+  securityContext: {}
   agentSidecar:
     enabled: true
   annotations: {}
@@ -360,6 +368,8 @@ spark:
 
 hotrod:
   enabled: false
+  podSecurityContext: {}
+  securityContext: {}
   replicaCount: 1
   image:
     repository: jaegertracing/example-hotrod
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.16.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
-version: 0.19.0
+version: 0.19.1
 keywords:
   - jaeger
   - opentracing
--- a/charts/jaeger/README.md
+++ b/charts/jaeger/README.md
@@ -185,6 +185,8 @@ The following table lists the configurable parameters of the Jaeger chart and th
 | `<component>.nodeSelector` | Node selector | {} |
 | `<component>.tolerations` | Node tolerations | [] |
 | `<component.affinity` | Affinity | {} |
+| `<component>.podSecurityContext` | Pod security context | {} |
+| `<component>.securityContext` | Container security context | {} |
 | `agent.annotations` | Annotations for Agent | `nil` |
 | `agent.cmdlineParams` |Additional command line parameters| `nil` |
 | `agent.dnsPolicy` | Configure DNS policy for agents | `ClusterFirst` |
--- a/charts/jaeger/templates/agent-ds.yaml
+++ b/charts/jaeger/templates/agent-ds.yaml
@@ -28,6 +28,8 @@ spec:
 {{ toYaml .Values.agent.podLabels | indent 8 }}
 {{- end }}
     spec:
+      securityContext:
+{{ toYaml .Values.agent.podSecurityContext | indent 8 }}
       {{- if .Values.agent.useHostNetwork }}
       hostNetwork: true
       {{- end }}
@@ -35,6 +37,8 @@ spec:
       serviceAccountName: {{ template "jaeger.agent.serviceAccountName" . }}
       containers:
       - name: {{ template "jaeger.agent.name" . }}
+        securityContext:
+{{ toYaml .Values.agent.securityContext | indent 10 }}
         image: {{ .Values.agent.image }}:{{ .Values.tag }}
         imagePullPolicy: {{ .Values.agent.pullPolicy }}
         env:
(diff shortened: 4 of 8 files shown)


mehta-ankit:
@naseemkullah @pavelnikolov @dvonthenen Can i get a . review on this please.

charts/jaeger/values.yaml
```diff
@@ -11,6 +11,12 @@ tag: 1.16.0
 nameOverride: ""
 fullnameOverride: ""
 
+# Security Context for deployments
+securityContext:
```
naseemkullah: Can we have  both podSecurityContext as well as container level security context? As done in `helm create`: https://github.com/helm/helm/blob/a963736f6675e972448bf7a5fd141628fd0ae4df/pkg/chartutil/create.go#L112-L121

naseemkullah: Can we have both podSecurityContext as well as container level security context? As done in helm create: https://github.com/helm/helm/blob/a963736f6675e972448bf7a5fd141628fd0ae4df/pkg/chartutil/create.go#L112-L121

charts/jaeger/values.yaml
```diff
@@ -11,6 +11,12 @@ tag: 1.16.0
...
+  runAsUser: 1000
+  runAsGroup: 3000
+  fsGroup: 2000
```
naseemkullah: Any reason in particular why `fsGroup` and `runAsGroup` differ?

charts/jaeger/values.yaml
mehta-ankit: @naseemkullah Sure i can add both. Also are you ok with default values i provided in Values.yaml (wanted to confirm) ??

mehta-ankit: I added podSecurityContext for pods and SecurityContext for the containers and made default values empty so as to not break current users use case.

naseemkullah: This is looking really good! Not sure on breakage/defaults as I have not tested, but allowing the user to is definitely a safe first step, good idea.

But I think we might want to nest the `securityContext` and `podSecurityContext` values per component (.e.g `.Values.collector.podSecurityContext`), the same way they each have their own affinity, tolerations and so on... but maybe I'm overthinking it. WDYT?

mehta-ankit: IMO, usually in production one would not want their containers or pods to run with root privileges.Also as defined here, without giving root privileges one can give certain capabilities to their containers.
https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-capabilities-for-a-container
Since we kept SecurityContext its own object, one can add this key as an override and it should work.

For jaeger's components i don't see why collector or query would need different context than say an agent. ES or Cassandra can have different context anyways (by passing values to their subchart). So i don't see its need.

Let me what you think @naseemkullah ?

naseemkullah: Hmm I think we should test: If we wanted to give tightest securtiy context possible, can we try:

```yaml
podSecurityContext: {}
   fsGroup: 2000
securityContext: {}
   capabilities:
   drop:
   - ALL
  readOnlyRootFilesystem: true
  runAsNonRoot: true
  runAsUser: 1000
```

If some workloads run properly but others do not, it might be a sign that each component having its own security context values would be a good idea.

matthiaslee: @mehta-ankit, I agree, as discussed f2f, I think it's a good idea to have these security contexts attached to each of the components, therefore if someone wants to be more restrictive with one component over another, they can. Additionally, its more explicit and easier to understand exactly what the security context will be attached to. For example we do not support attaching the security context to ES/Cassanda, if the security context is defined at the root level, it's ambiguous which components it is attached to.

mehta-ankit: @naseemkullah Made the change as discussed above. Can you review it again please. 👍  Thanks.

naseemkullah: lgtm!

naseemkullah:
Please gpg sign all you commits @mehta-ankit

charts/jaeger/values.yaml
naseemkullah: In fact let's go ahead and add as default value for all components' securityContext:

```
  runAsNonRoot: true
  runAsUser: 1000
```

That should be fine and seems like a sane default. If you would mind testing first that would be swell. We can maybe add more security tightening in a future PR but for now this would be good.

charts/jaeger/values.yaml
```diff
@@ -306,6 +318,10 @@ spark:
 # End: Default values for the various components of Jaeger
 
 hotrod:
+  # pod level security context
```
naseemkullah: Can we move these comments to the README. you can make one line like 
```
<component>.podSecurityContext | Pod security context | `nil`
<component>.securityContext | Container security context | `runAsNonRoot: true`<br>`runAsUser: 1000`
```

mehta-ankit: Sure. Can do that.

charts/jaeger/values.yaml
mehta-ankit: Do we need both ?? If we say runAsUser 1000 its a non root user anyways (container would run with spcific uid). On the other hand if we say `runAsNonRoot: true`, why do we care about the uid, when it would be a non root user's uid. 🤔  Do you see my point here. ?
Also, as i said in my comment [here](https://github.com/jaegertracing/helm-charts/pull/37#discussion_r370105485) do we want to tighten the security by default, instead of giving users of jaegers chart a chance to provide values as they see fit for their use case.

mehta-ankit: Also, in case you still want the defaults as mentioned in your [comment](https://github.com/jaegertracing/helm-charts/pull/37#discussion_r370822935) do we want these for both containerSecurityContext and podSecurityContext sections for all the components.??

naseemkullah: > Do we need both ?? If we say runAsUser 1000 its a non root user anyways (container would run with spcific uid). On the other hand if we say runAsNonRoot: true, why do we care about the uid, when it would be a non root user's uid. 🤔 Do you see my point here. ?

Yeah I do see what you mean, if we specify runAsUser, the runAsNonRoot might be redundant. I'm not familiar with best practices, I generally put both because I originally used https://kubesec.io/ to audit my workloads and it suggested to add both 🤷‍♂ ... But I see in another audit tool called Polaris, both would not be needed https://github.com/FairwindsOps/polaris/blob/91378c3d920d8ac79b197cc44c171e50ac953d32/CHANGELOG.md#051

So yeah probably just one of the two would suffice!

> Also, in case you still want the defaults as mentioned in your [comment](https://github.com/jaegertracing/helm-charts/pull/37#discussion_r370822935) do we want these for both containerSecurityContext and podSecurityContext sections for all the components.??

I'd put those in just the container `securityContext`  of all components, but have not but excessive thought into this. 

But the idea would be to run something like Polaris and reduce the number of warnings by default.

naseemkullah: Sorry for the back and forth on this, but let's leave them blank for now to get this PR in, and work on setting sane defaults in a separate PR.

mehta-ankit: Sounds Good to me. I like the discussion on this though. I can make the other PR for sane defaults and we can continue the discussion there.
Will push the README changes and sign both commits and you can approve the PR then.

mehta-ankit:
@naseemkullah I rebased to resolve conflicts and updated the version to 0.19.1.
Can you approve and merge please. 🙏 👍

naseemkullah:
Sure but we added a new component "infester" could you add values for it too please?

mehta-ankit:
> Sure but we added a new component "ingester" could you add values for it please?

Sure. Added it to the ingester.

naseemkullah:
Great, thanks!

PR #43, opened 2020-02-03
[jaeger] Update readme

naseemkullah:
- Lint as per https://github.com/DavidAnson/markdownlint/blob/v0.18.0/doc/Rules.md
- Fix service account values
- Update recently added images to use ones from jaeger website


arpitjindal97:
This also looks good to me. We can merge this

PR #46, opened 2020-02-14
Add elasticsearch tls support to spark

cgroschupp:


charts/jaeger/README.md
```diff
@@ -223,137 +225,139 @@ helm install jaeger jaegertracing/jaeger \
...
-| `storage.elasticsearch.port` | Provisioned elasticsearch port| `9200` |
-| `storage.elasticsearch.scheme` | Provisioned elasticsearch scheme | `http` |
-| `storage.elasticsearch.usePassword` | Use password | `true` |
-| `storage.elasticsearch.user` | Provisioned elasticsearch user| `elastic` |
-| `storage.elasticsearch.nodesWanOnly` | Only access specified es host | `false` |
-| `storage.kafka.brokers` | Broker List for Kafka with port | `kafka:9092` |
-| `storage.kafka.topic` | Topic name for Kafka | `jaeger_v1_test` |
-| `storage.type` | Storage type (ES or Cassandra)| `cassandra` |
-| `tag` | Image tag/version | `1.16.0` |
+| Parameter                                                 | Description                                                                                                                                                      | Default                                 |
```
naseemkullah: Could we keep the spacing as it was before?
It is more readable in an IDE imho... BTW how did you space it all out?

charts/jaeger/README.md
```diff
@@ -170,24 +170,26 @@ collector:
...
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: jaeger-tls
-data:
-  es.pem: |
-    -----BEGIN CERTIFICATE-----
-    <CERT>
-    -----END CERTIFICATE-----
+Generate configmap jaeger-tls:
```
naseemkullah: Is it not better to put this data in a secret rather than a cm?

cgroschupp: For my use case the configmap is ok, because it is only the certificate, not the private key.

cgroschupp: But with extraSecretMounts it is also possible to use a secret.

charts/jaeger/README.md
cgroschupp: My IDE formats the markdown automatically.

naseemkullah: Oh I see, thanks.

cgroschupp: i have disabled the table autoformatting feature of my IDE and committed the changes

naseemkullah: Thanks! Hmmm maybe I should turn on mine. :)
 I will look into that but thanks for the info and reverting the change.

PR #45, opened 2020-02-07
Added cassandra auth variables in spark job

arpitjindal97:
- cassandra auth variables were missing in spark job
    cmdlineParams were also missing

 - fixed section {{- define "elasticsearch.client.url" -}} in helpers
    it was generating wrong url

PR #47, opened 2020-02-14
Allow to use external es host

cgroschupp:
After #45 it is no longer possible to set an external elasticsearch host.

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.16.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
-version: 0.19.3
+version: 0.19.5
 keywords:
   - jaeger
   - opentracing
--- a/charts/jaeger/templates/_helpers.tpl
+++ b/charts/jaeger/templates/_helpers.tpl
@@ -213,19 +213,13 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 */}}
 {{- define "elasticsearch.client.url" -}}
 {{- $port := .Values.storage.elasticsearch.port | toString -}}
-{{- if .Values.provisionDataStore.elasticsearch -}}
-{{- if .Values.storage.elasticsearch.nameOverride }}
-{{- $host := printf "%s" .Values.storage.elasticsearch.nameOverride | trunc 63 | trimSuffix "-" -}}
-{{- printf "%s://%s:%s" .Values.storage.elasticsearch.scheme $host $port }}
-{{- else }}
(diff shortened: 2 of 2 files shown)


naseemkullah:
@arpitjindal97 You recently made this change and @cgroschupp would like to change it back

Could you both please explain how this helper is affecting the es host name?

arpitjindal97:
Hi everyone, please give me some hours.

I'm currently in an event in SAP. I don't know how it is breaking. Just give me some hours. Let me check it once.

cgroschupp:
my values:
```
storage:
  type: elasticsearch
  elasticsearch:
    host: myhost
    scheme: https
provisionDataStore:
  elasticsearch: false
```
The if block of the helper returns no value if provisionDataStore.elasticsearch is set to false.
But this is necessary when the es not managed by the chart.

naseemkullah:
I see

Would it make sense to simply have

`{{- printf "%s://%s:%s" .Values.storage.elasticsearch.scheme .Values.storage.elasticsearch.host $port }}`

regardless of provisioned or not? Please both let me know if that solves use cases of provisioned and external ES.

arpitjindal97:
#49 fixed everything

cgroschupp:
we have three cases:
* provisionDataStore=false
* provisionDataStore=true
* provisionDataStore=true and nameOverride

I do not know, but I think this is easier to understand:
```
{{- define "elasticsearch.client.url" -}}
{{- $port := .Values.storage.elasticsearch.port | toString -}}
{{- $host := .Values.storage.elasticsearch.host -}}
{{- if .Values.provisionDataStore.elasticsearch }}
{{- $host := printf "%s-%s-%s" .Release.Name "elasticsearch" "client" | trunc 63 | trimSuffix "-" -}}
{{- if .Values.storage.elasticsearch.nameOverride }}
{{- $host := printf "%s" .Values.storage.elasticsearch.nameOverride | trunc 63 | trimSuffix "-" -}}
{{- end -}}
{{- end }}
{{- printf "%s://%s:%s" .Values.storage.elasticsearch.scheme $host $port }}
{{- end -}}
```

cgroschupp:
> #49 fixed everything

see: https://github.com/jaegertracing/helm-charts/pull/49#issuecomment-586320169

cgroschupp:
> I see
> 
> Would it make sense to simply have
> 
> `{{- printf "%s://%s:%s" .Values.storage.elasticsearch.scheme .Values.storage.elasticsearch.host $port }}`
> 
> regardless of provisioned or not? Please both let me know if that solves use cases of provisioned and external ES.

@naseemkullah You're right, I changed it.

naseemkullah:
lgtm @arpitjindal97  does this work for you?

Makes me think we should just introduce storage.elasticsearch.url in a non-breaking mannger... but that could be a separate PR

arpitjindal97:
Yes, It looks good to me. I think it will work for all scenarios. We can merge it.

PR #49, opened 2020-02-14
fixed breaking url issue

arpitjindal97:


cgroschupp:
I think {{- $host := printf "%s-%s-%s" .Release.Name "elasticsearch" "client" | trunc 63 | trimSuffix "-" -}}` is missing, for an es managed by the chart.

arpitjindal97:
This line was the initial reason why this problem arised and I made PR to fix it. 
ElasticSearch service has a different name and this line output a different URL.

Components were never able to communicate to ES

cgroschupp:
The default service name has been changed, see:
https://github.com/elastic/helm-charts/blob/7.5.1/elasticsearch/templates/service.yaml#L5
https://github.com/elastic/helm-charts/blob/7.5.1/elasticsearch/templates/_helpers.tpl#L19

And the defaults are:
```
nodeGroup: master
clusterName: elasticsearch
```

We need only to change `client` to `master` for a default installation.

arpitjindal97:
We have put elasticsearch-master in values as host. So, instead of hardcoding in template. We can use host variable in template

naseemkullah:
Can we close this @arpitjindal97 ?

PR #51, opened 2020-02-20
Added Elasticsearch indexPrefix option

arpitjindal97:
Similar to keyspace in cassandra


naseemkullah:
Thanks @arpitjindal97!

I think it would be best to wrap these around if statements in the templates, and by default not set any aka comment out the `indexPrefix` line in values.yaml (or set its' value `null`? I forget if that works).

naseemkullah:
Oh and please add the value in README for this optional prefix of Jaeger indices. 🙏

arpitjindal97:
hahaha, Sure

naseemkullah:
👍

PR #54, opened 2020-02-24
Fixed cassandra-tls secret mount issue

arpitjindal97:


naseemkullah:
Could this be fixed by retaining the auto mounting of secret just by setting `cassandra.tls.enabled` and `storage.cassandra.secretName` I believe this will result in a much better user experience. We are probably just missing something small 🤔

arpitjindal97:
We can do that also but I removed the auto mounting part. because then question comes of automounting Kafka TLS and elasticsearch TLS

naseemkullah:
Good point. I think in general it is a more "helmy" experience to just say "yes enable, and here is my secret name" and then the chart does the hard work. extra* type stuff is for specific cases, but i think enabling tls is a common case. Makes sense?

naseemkullah:
@pavelnikolov @dvonthenen Could one of you kindly take this issue (I'm a little swamped at the moment)?

naseemkullah:
Think it would be a good idea to add `cqlshrc ` as a configmap in our templates and automatically mount it when cassandra tls is enabled? I think it makes more sense rather than ask the user to add this in a secret.

pavelnikolov:
LGTM. However, I've never used the chart with Cassandara (only with Elasticsearch). That is why I'm not the best person to review this PR.

pavelnikolov:
I agree with @naseemkullah that a single secret is enough. I think that you don't even need `enabled` if the secret name is set, then it can be considered as enabled. However, we need to document the expected paths in the secret.

PR #55, opened 2020-02-26
Bump jaeger to 1.17.0

naseemkullah:
Signed-off-by: Naseem <naseem@transit.app>

### DIFF
--- a/charts/jaeger-operator/Chart.yaml
+++ b/charts/jaeger-operator/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
 description: jaeger-operator Helm chart for Kubernetes
 name: jaeger-operator
-version: 2.13.1
-appVersion: 1.16.0
+version: 2.14.0
+appVersion: 1.17.0
 home: https://www.jaegertracing.io/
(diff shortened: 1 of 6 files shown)


naseemkullah:
> LGTM. The Operator will be released soon as well. Would you prefer to have both bumps as part of different commits, or would you prefer to wait for the Operator bump before merging this?

Thanks for pointing that out @jpkrohling, I updated operator to 1.17 assuming that it all released at the same time. So I will wait to merge as you've mentioned it will be released soon.

jpkrohling:
The Jaeger Operator v1.17.0 has just been released. Note that we found a breaking change affecting users of Kafka as the storage backend and/or ingester when TLS is enabled. Not sure there's anything you can do in the Helm side, but you might want to take a look at our upgrade mechanisms in the operator, to see what we are doing there to address this.

This is the issue description: https://github.com/jaegertracing/jaeger/pull/1838#discussion_r385020296

And this is the upgrade code we have:
https://github.com/jaegertracing/jaeger-operator/blob/c9147e7ec923113b2dbeafc917c8eb0070c7dc8e/pkg/upgrade/v1_17_0.go#L51-L67

PR #62, opened 2020-03-03
[ci] Use actions checkout v2

naseemkullah:
We've been experiencing https://github.com/actions/checkout/issues/23 quite a bit lately. v2 should fix that.


naseemkullah:
Any idea as to why this is failing the lint-and-test step if no chart changes are being made?
cc @scottrigby

PR #63, opened 2020-03-03
Add service accoiunt for es-index-cleaner object

mehta-ankit:

Chart version 0.22.0 has a bug:

If `.Values.esIndexCleaner.serviceAccount.create` is set to true to create a service account, the cleaner job fails to create a pod because no service account is created (as there is no template for the service account).

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.17.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
-version: 0.22.0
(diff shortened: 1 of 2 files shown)


mehta-ankit:
@naseemkullah Forgot to add the service account for `es-index-cleaner` cronjob. 🤦‍♂

mehta-ankit:
i also feel, i should have (and others can do this as well) pushed my changes with a commit enabling the es-index-cleaner (or any new component we create). This would have caught the issue in CI.

naseemkullah:
Thanks @mehta-ankit , yes our test suite is not very thorough.

PR #67, opened 2020-03-06
use "enabled" key to add zipkin port to collector

golgoth31:
use "enabled" key to add zipkin port to collector (avoid warning from helm)

Currently, only the key "zipkin" is used and helm complains about it:
2020/03/06 09:55:11 Warning: Merging destination map for chart 'jaeger'. The destination item 'zipkin' is a table and ignoring the source 'zipkin' as it has a non-table value of: <nil>
2020/03/06 09:55:11 Warning: Merging destination map for chart 'jaeger'. The destination item 'zipkin' is a table and ignoring the source 'zipkin' as it has a non-table value of: <nil>

Signed-off-by: David Sabatie <david.sabatie@delair.aero>

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.17.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
-version: 0.22.1
+version: 0.22.2
 keywords:
   - jaeger
   - opentracing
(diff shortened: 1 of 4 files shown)


golgoth31:
Hi,
Is there any way to make the lint-test ok ?
thanks

naseemkullah:
Hi @golgoth31 thanks for bringing this up.
A nitpick: could we try adding empty braces as follows? I prefer this to having only one of the ports with an `enabled` value but can resort to that if below does not work.

```
    zipkin: {}
      # port: 9411
```

golgoth31:
No problem, I make the test

golgoth31:
Done, it's working as expected, without warnings. Thank you

naseemkullah:
Thanks @golgoth31!

Issue #74, opened 2020-03-12
[jaeger] Give user more unlimited config options for storage


naseemkullah:
https://github.com/jaegertracing/helm-charts/blob/2f60c580bd4729b18a3173d7a44f8eeafafc481d/charts/jaeger/values.yaml#L18-L44 is not scalable as additional config options get added/removed from upstream jaeger.

It would be great to be able to provide any config options by a variety of means the user can choose from (any or a mix of the following: env vars, env vars from secrets, env vars from configmaps, comman line options).

This has been prompted by consistent PRs such as https://github.com/jaegertracing/helm-charts/pull/73 and https://github.com/jaegertracing/helm-charts/pull/65 to bubble up config options.

vipinvkmenon:
Based on this particular line from the documentation “In order to provide configuration parameters via environment variables, find the respective command line option and convert its name to UPPER_SNAKE_CASE” and considering as more and more flags and features would be coming in.. would it be nice to provide a support on the chart to accept additional environment variables?

more configuration freedom as well☺️

naseemkullah:
💯 

We will allow additional env vars as well as command line options (user can take their pick).

naseemkullah:
Closed via #75

PR #75, opened 2020-03-13
refactor-config

naseemkullah:
* cmd line options are templated as command arguments as they should be.
* components now have an env value to configure any env vars.
* Cassandra/Elasticsearch values that are required in the various components that talk to them (collector/ingester, query, spark-job, cassandra-schema, es-index-cleaner, however the latter three have non-standard env var names that are kept) have been moved to _helpers.tpl for shareability.

### DIFF
--- a/charts/jaeger/templates/collector-deploy.yaml
+++ b/charts/jaeger/templates/collector-deploy.yaml
@@ -45,7 +45,11 @@ spec:
         imagePullPolicy: {{ .Values.collector.pullPolicy }}
         args:
           {{- range $key, $value := .Values.collector.cmdlineParams -}}
+          {{- if $value }}
           - --{{ $key }}={{ $value }}
+          {{- else }}
+          - --{{ $key }}
+          {{- end }}
           {{- end -}}
           {{- if not .Values.ingester.enabled -}}
           {{- include "storage.cmdArgs" . | nindent 10 }}
--- a/charts/jaeger/templates/collector-deploy.yaml
+++ b/charts/jaeger/templates/collector-deploy.yaml
@@ -43,11 +43,14 @@ spec:
 {{ toYaml .Values.collector.securityContext | indent 10 }}
         image: {{ .Values.collector.image }}:{{ .Values.tag }}
         imagePullPolicy: {{ .Values.collector.pullPolicy }}
-        env:
-          {{- range $key, $value := .Values.collector.cmdlineParams }}
-          - name: {{ $key | replace "." "_" | replace "-" "_" | upper | quote }}
-            value: {{ $value | quote }}
+        args:
+          {{- range $key, $value := .Values.collector.cmdlineParams -}}
+          - --{{ $key }}={{ $value }}
+          {{- end -}}
+          {{- if not .Values.ingester.enabled -}}
+          {{- include "storage.cmdArgs" . | nindent 10 }}
           {{- end }}
+        env:
           {{- if .Values.collector.service.zipkin }}
           - name: COLLECTOR_ZIPKIN_HTTP_PORT
             value: {{ .Values.collector.service.zipkin.port | quote }}
@@ -55,6 +58,10 @@ spec:
           {{- if .Values.ingester.enabled }}
           - name: SPAN_STORAGE_TYPE
             value: kafka
+          {{- range $key, $value := .Values.storage.kafka.env }}
+          - name: {{ $key | quote }}
+            value: {{ $value | quote }}
+          {{- end }}
           - name: KAFKA_PRODUCER_BROKERS
             value: {{ include "helm-toolkit.utils.joinListWithComma" .Values.storage.kafka.brokers }}
           - name: KAFKA_PRODUCER_TOPIC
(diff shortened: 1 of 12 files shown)


naseemkullah:
Fixes #74
cc @vipinvkmenon @arpitjindal97 @mehta-ankit  @cgroschupp @ssoerensen  Please take a look and share thoughts.

naseemkullah:
@pavelnikolov  @dvonthenen  there were some whitespace changes in helpers file, I'd suggest ignoring whitespace changes while reviewing: 
image


FWIW I deployed this version of the chart with default values as well as with ES per `storage.type=elasticsearch,provisionDataStore.cassandra=false,provisionDataStore.elasticsearch=true` and they both work.

charts/jaeger/templates/collector-deploy.yaml
```diff
@@ -43,74 +43,33 @@ spec:
 {{ toYaml .Values.collector.securityContext | indent 10 }}
         image: {{ .Values.collector.image }}:{{ .Values.tag }}
         imagePullPolicy: {{ .Values.collector.pullPolicy }}
-        env:
-          {{- range $key, $value := .Values.collector.cmdlineParams }}
-          - name: {{ $key | replace "." "_" | replace "-" "_" | upper | quote }}
-            value: {{ $value | quote }}
+        args:
+          {{- range $key, $value := .Values.collector.cmdlineParams -}}
+          - --{{ $key }}={{ $value }}
```
pavelnikolov: some command line params do not have a value, right? Maybe the `={{ $value }}` part should be optional

charts/jaeger/templates/ingester-deploy.yaml
```diff
@@ -47,64 +47,19 @@ spec:
 {{ toYaml .Values.ingester.securityContext | indent 10 }}
         image: {{ .Values.ingester.image }}:{{ .Values.tag }}
         imagePullPolicy: {{ .Values.ingester.pullPolicy }}
-        env:
+        args:
           {{- range $key, $value := .Values.ingester.cmdlineParams }}
-          - name: {{ $key | replace "." "_" | replace "-" "_" | upper | quote }}
-            value: {{ $value | quote }}
+          - --{{ $key }}={{ $value }}
```
pavelnikolov: same as above

naseemkullah:
Reverting to actions/checkout@v1 from actions/checkout@v2 in CI, it seems like CI is broken with v2

Will stick with v1 until https://github.com/helm/chart-testing-action#example-workflow is updated with v2

charts/jaeger/templates/collector-deploy.yaml
naseemkullah: I am having trouble finding an example of this: https://www.jaegertracing.io/docs/1.17/cli/#jaeger-collector

Could you point one out?

charts/jaeger/templates/ingester-deploy.yaml
naseemkullah: same as above :p https://www.jaegertracing.io/docs/1.17/cli/#jaeger-ingester

FWIW even --help seems to be able to take true/false

charts/jaeger/templates/collector-deploy.yaml
naseemkullah: @jpkrohling @yurishkuro Could you please confirm if any command line options for any of the components are in the form of `--key` without a value (i.e. `--key=value`) ?

pavelnikolov: AFAIK, boolean flags only need a key...

charts/jaeger/templates/ingester-deploy.yaml
pavelnikolov: But think about it when you run the executable you only ever provide `--help` and it changes the value from `false` to `true`. It makes sense for all boolean flags to behave this way

naseemkullah: So we have two choices:

* have users explicitly set true for the boolean flags they wish to enable
or
* templating trickery to detect if a value was provided with the key only for boolean flags that are disabled by default

naseemkullah: tbh I'm not sure how to achieve the latter. Any suggestions?

pavelnikolov: The easiest possible option that comes to my mind is treating the entire flag as one string. Then it's just a string iteration:

```
--some-bool-flag
--some-int-flag=5
```

i.e. don't split by key/value.

pavelnikolov: or maybe conditionally print `={{ $value }}` if the value is not nil.

naseemkullah: Found an example frmo a chart in the stable repo:

```
          {{- if $value }}
            - --{{ $key }}={{ $value }}
          {{- else }}
            - --{{ $key }}
          {{- end }}
```

Will try that!

naseemkullah: The above works with the caveat that the key only value must be defined with a trailing `:` as such:

```
    cmdlineParams:
      foo:
      bar: baz
```

which renders:

```
          - --foo
          - --bar=baz
```

Good enough? 😬

charts/jaeger/templates/collector-deploy.yaml
naseemkullah: Alright, please see https://github.com/jaegertracing/helm-charts/pull/75/commits/8d36352e78b758e2fd7eb4d3006713a16a4a8366

charts/jaeger/templates/ingester-deploy.yaml
naseemkullah: Gonna suppose so 😅 here's https://github.com/jaegertracing/helm-charts/pull/75/commits/8d36352e78b758e2fd7eb4d3006713a16a4a8366

naseemkullah: https://github.com/jaegertracing/helm-charts/pull/75/commits/64352aea620b02d8cb8c13ae786df842af8348c3

Voilà!

charts/jaeger/templates/collector-deploy.yaml
naseemkullah: Voilà!

https://github.com/jaegertracing/helm-charts/pull/75/commits/64352aea620b02d8cb8c13ae786df842af8348c3

PR #82, opened 2020-03-16
[jaeger]Use nindent over indent

naseemkullah:
Use nindent function and indent the templating pipelines where the values will be, instead of indent function without indenting the templating pipelines to improve readability.


naseemkullah:
Merging to also test release workflow.

PR #87, opened 2020-03-26
Added support for specifying the port number to use for the node port  for the Jaeger query service

marcusportmann:
Currently a random node port number is generated for the Jaeger query service when using the NodePort service type. 

This pull request adds support for specifying the port number to use for the node port  for the Jaeger query service.


naseemkullah:
Thanks @marcusportmann, I'm investigating the build breakage but LGTM

PR #95, opened 2020-04-08
Allow setting of agent pod priorityClassName

lheinlen:
We do a lot of autoscaling in our cluster and have run into issues where the Jaeger agent pods end up stuck in `Pending` because other pods with the default priority filled the node before the agent pod could be scheduled. To prevent this, we added the ability to set the `priorityClassName` on the agent pods.


naseemkullah:
Thanks @lheinlen, lgtm

lheinlen:
Thank you!

PR #109, opened 2020-05-21
Added config to make operator watch in arbitrary namespaces

davelosert:
The current helm-chart only allows a somewhat "all or nothing" strategy when it comes to RBAC and watching namespaces: Either the operator only has rights and watches the namespace it was deployed to or it has cluster-wide permissions with a `ClusterRoleBinding`.

This PR introduces a configuration to watch for arbitrary namespaces.
If set, it also creates a `ClusterRole`, but only `RoleBindings` in all defined namespaces.


davelosert:
Setting `WATCH_NAMESPACE` to a comma-separated list currently does not work with the operator due to a bug I reported in the [operator-repository](https://github.com/jaegertracing/jaeger-operator/issues/1069)

So I guess this PR should only be merged once that bug got fixed.

davelosert:
@cpanato : Updated the PR and adjusted the version. Should be ready for merge now 🙂

cpanato (approved):
/lgtm

rafaribe:
What's keeping this PR from getting merged? I'm currently experiencing an issue with the namespaces and I think this could fix it.

DodgeCamaro:
Any updates?

stevehipwell:
I'm not sure the logic here is correct, nor is it correct currently either, as setting `jaeger.namespace` should override the release namespace. If this extra logic is actually required at that point it should be layered on top of this making sure that the CR created by the chart can always be observed.

PR #110, opened 2020-05-28
Add custom SecretKey Option for Elasticsearch

oliverbaehler:
I was deploying Jeager with an Elasticsearch Backend. But I didn't use the elasticsearch Subchart but a custom chart, which supports the creation of Elastic related crds provided by the ECK Operator (https://github.com/elastic/cloud-on-k8s). When deploying an elasticsearch cluster via eck-operator there will be a secret created with the credentials for the elastic user. The difference is, that the credentials are stored under the .elastic key in the secret. To resolve this issue I added an Option to define the key within the referenced secret.

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.17.1
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
-version: 0.28.0
+version: 0.29.0
 keywords:
   - jaeger
   - opentracing
--- a/charts/jaeger/README.md
+++ b/charts/jaeger/README.md
@@ -346,6 +346,7 @@ The following table lists the configurable parameters of the Jaeger chart and th
 | `storage.elasticsearch.env` | Extra ES related env vars to be configured on components that talk to ES | `nil` |
 | `storage.elasticsearch.cmdlineParams` | Extra ES related command line options to be configured on components that talk to ES | `nil` |
 | `storage.elasticsearch.existingSecret` | Name of existing password secret object (for password authentication | `nil` |
+| `storage.elasticsearch.existingSecretKey` | Key of the declared password secret | `password` |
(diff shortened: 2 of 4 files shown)


naseemkullah:
Thanks @oliverbaehler, all commits require to be gpg signed, could you please gpg sign your commit?

naseemkullah:
https://help.github.com/en/github/authenticating-to-github/about-commit-signature-verification for more info.

oliverbaehler:
@naseemkullah Should be okay now =)

oliverbaehler:
Another question regarding Elasticsearch (no directly connected to this pr). Is there any other purpose to the jaegertracing/jaeger-es-index-cleaner than clearing indices after a certain amount of time? I just wanna point out, that this funcionality would be easily replaceable by Elasticsearch Lifecycle Policies (https://www.elastic.co/guide/en/kibana/current/index-lifecycle-policies.html). They serve the purpose to manage the lifecycle of indices (based on pattern) so you could define the lifespan of the jeager indicies within such a policy, might reduce the complexity of the chart and is just a thought.

naseemkullah:
> Another question regarding Elasticsearch (no directly connected to this pr). Is there any other purpose to the jaegertracing/jaeger-es-index-cleaner than clearing indices after a certain amount of time? I just wanna point out, that this funcionality would be easily replaceable by Elasticsearch Lifecycle Policies (https://www.elastic.co/guide/en/kibana/current/index-lifecycle-policies.html). They serve the purpose to manage the lifecycle of indices (based on pattern) so you could define the lifespan of the jeager indicies within such a policy, might reduce the complexity of the chart and is just a thought.

That makes a lot of sense, we could go with ILM or even elastic curator, but @pavolloffay and the team decided it is best to wrap it (es curator) as `jaeger-es-index-cleaner`, as long as they continue to maintain this convenience wrapper I'd recommend it as the means, but nothing wrong to manage ILM/es-curator yourself if you prefer that.

PR #119, opened 2020-06-10
allow to set env variables for spark cronjob

cgroschupp:


naseemkullah:
Thanks @cgroschupp ... could you please add `env: {}` somewhere here: https://github.com/jaegertracing/helm-charts/blob/0dc06f9b5fd9da485ae91477400e84a84f535481/charts/jaeger/values.yaml#L391-L418 as well?

naseemkullah:
Hi @cgroschupp 

I believe for nothing but consistency across the chart sake, we should stick to the following format for env vars:

https://github.com/jaegertracing/helm-charts/blob/6787ff09cec9d06290f999d5a5d3fb64edd6ba77/charts/jaeger/templates/cassandra-schema-job.yaml#L34-L37


if from configmaps and secrets we use the corresponding mount values

cgroschupp:
@naseemkullah This format is not flexible enough. It is not possible to use "valueFrom" for Secret or ConfigMap.

cgroschupp:
This format is not flexible enough. It is not possible to use "valueFrom" for values from Secret or ConfigMap. How it is possible to set env variables with volumeMounts?

naseemkullah:
> This format is not flexible enough. It is not possible to use "valueFrom" for values from Secret or ConfigMap. How it is possible to set env variables with volumeMounts?

You are right @cgroschupp ... For consistency sake I think this should be done for evert component but will be a major breaking change. I've created https://github.com/jaegertracing/helm-charts/issues/120 in case we can think of other breaking changes to introduce when we do this. WDYT?

For now I approved adding this to spark cron job in case this is a blocker for you.

cgroschupp:
@naseemkullah I also think it is best to change all it to `env: []`. Only Values.esIndexCleaner.env and .Values.schema.env are affected

charts/jaeger/templates/spark-cronjob.yaml
```diff
@@ -42,6 +42,7 @@ spec:
               - name: STORAGE
                 value: {{ .Values.storage.type }}
               {{- include "storage.env" . | nindent 14 }}
+              {{- toYaml .Values.spark.env | nindent 14 }}
```
naseemkullah, line 45:
```suggestion
              {{- toYaml .Values.spark.extraEnv | nindent 14 }}
```
opening this suggestion as per https://github.com/jaegertracing/helm-charts/issues/120#issuecomment-643047774 which is still open to discussion of course :)

charts/jaeger/values.yaml
```diff
@@ -395,6 +395,7 @@ spark:
   tag: latest
   pullPolicy: Always
   cmdlineParams: {}
+  env: []
```
naseemkullah, line 398:
```suggestion
  extraEnv: []
```

opening this suggestion as per https://github.com/jaegertracing/helm-charts/issues/120#issuecomment-643047774 which is still open to discussion of course :)

cgroschupp: @naseemkullah done

charts/jaeger/README.md
```diff
@@ -313,6 +313,7 @@ The following table lists the configurable parameters of the Jaeger chart and th
 | `schema.activeDeadlineSeconds` | Deadline in seconds for cassandra schema creation job to complete | `120` |
 | `schema.keyspace` | Set explicit keyspace name | `nil` |
 | `spark.enabled` | Enables the dependencies job| `false` |
+| `spark.extraEnv` | Additional environment variables | {} |
```
mehta-ankit, line 316: @naseemkullah @cgroschupp I think there is a typo, `{}` should be `[]`, as that's what you set as default in [Values.yaml file](https://github.com/jaegertracing/helm-charts/pull/119/files#diff-630d916c964d14d3d12a7a96261e789fR398).

naseemkullah: good catch @mehta-ankit , this will be addressed in https://github.com/jaegertracing/helm-charts/pull/121, please take a look!

PR #121, opened 2020-06-12
[jaeger] use extraEnv in favor of env

### LINKED ISSUE #120: [jaeger] increase env var flexibility
We should improve env var flexibility.

- change `env: {}` key value pairs to the more flexible `env: []` as-is yaml
  - see https://github.com/jaegertracing/helm-charts/pull/119 for more details

cc @pavelnikolov  @cgroschupp @mehta-ankit @Jasstkn  @arpitjindal97

### LINKED ISSUE #124: Upgrading to 0.30.1 causes an error
I tried using the chart version 0.30.1, and it throws an error:
```
Error: YAML parse error on /charts/jaeger/templates/spark-cronjob.yaml: error converting YAML to JSON: yaml: line 44: could not find expected ':'
 helm.go:75: [debug] error converting YAML to JSON: yaml: line 44: could not find expected ':'
```
I think its related to: https://github.com/jaegertracing/helm-charts/pull/119
 cc @naseemkullah

naseemkullah:
fixes #120
fixes #124 

env was limited in its implementation, extraEnv takes a list and performs a toYaml to enable maxiumum flexibility such as values from secrets and configmaps

cc @cgroschupp @Jasstkn can you please review if you get a chance?

### DIFF
--- a/charts/jaeger/templates/agent-ds.yaml
+++ b/charts/jaeger/templates/agent-ds.yaml
@@ -53,9 +53,12 @@ spec:
           {{- end }}
           {{- end }}
         env:
+        {{- if .Values.agent.extraEnv }}
+          {{- toYaml .Values.agent.extraEnv | nindent 10 }}
+        {{- end }}
         {{- if not (hasKey .Values.agent.cmdlineParams "reporter.grpc.host-port") }}
-        - name: REPORTER_GRPC_HOST_PORT
-          value: {{ include "jaeger.collector.name" . }}:{{ .Values.collector.service.grpc.port }}
+          - name: REPORTER_GRPC_HOST_PORT
+            value: {{ include "jaeger.collector.name" . }}:{{ .Values.collector.service.grpc.port }}
         {{- end }}
         ports:
         - name: zipkin-compact
--- a/charts/jaeger/templates/collector-deploy.yaml
+++ b/charts/jaeger/templates/collector-deploy.yaml
@@ -66,6 +66,9 @@ spec:
           - name: {{ $key | quote }}
             value: {{ $value | quote }}
           {{- end }}
+        {{- if .Values.storage.kafka.extraEnv }}
+          {{- toYaml .Values.storage.kafka.extraEnv | nindent 10 }}
+        {{- end }}
           - name: KAFKA_PRODUCER_BROKERS
             value: {{ include "helm-toolkit.utils.joinListWithComma" .Values.storage.kafka.brokers }}
           - name: KAFKA_PRODUCER_TOPIC
--- a/charts/jaeger/templates/es-index-cleaner-cronjob.yaml
+++ b/charts/jaeger/templates/es-index-cleaner-cronjob.yaml
@@ -41,13 +41,12 @@ spec:
             image: "{{ .Values.esIndexCleaner.image }}:{{ .Values.esIndexCleaner.tag }}"
             imagePullPolicy: {{ .Values.esIndexCleaner.pullPolicy }}
             args:
-            - {{ .Values.esIndexCleaner.numberOfDays | quote }}
-            - {{ include "elasticsearch.client.url" . }}
+              - {{ .Values.esIndexCleaner.numberOfDays | quote }}
+              - {{ include "elasticsearch.client.url" . }}
             env:
-              {{- range $key, $value := .Values.esIndexCleaner.env }}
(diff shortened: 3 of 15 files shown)


charts/jaeger/templates/es-index-cleaner-cronjob.yaml
```diff
@@ -44,10 +44,7 @@ spec:
             - {{ .Values.esIndexCleaner.numberOfDays | quote }}
             - {{ include "elasticsearch.client.url" . }}
             env:
-              {{- range $key, $value := .Values.esIndexCleaner.env }}
-              - name: {{ $key | quote }}
-                value: {{ $value | quote }}
-              {{- end }}
```
naseemkullah, line 50: This was not advertised in the values file, also it is non-critical so I am fine with removing it in this PR.

charts/jaeger/templates/agent-ds.yaml
```diff
@@ -53,10 +53,13 @@ spec:
           {{- end }}
           {{- end }}
         env:
-        {{- if not (hasKey .Values.agent.cmdlineParams "reporter.grpc.host-port") }}
-        - name: REPORTER_GRPC_HOST_PORT
-          value: {{ include "jaeger.collector.name" . }}:{{ .Values.collector.service.grpc.port }}
+        {{- if .Values.agent.extraEnv }}
+          {{- toYaml .Values.agent.extraEnv | nindent 10 }}
         {{- end }}
+          {{- if not (hasKey .Values.agent.cmdlineParams "reporter.grpc.host-port") }}
```
mehta-ankit, line 59: (nit)
I am not 100% sure, but i feel we should move this line `{{- if not (hasKey .Values.agent.cmdlineParams "reporter.grpc.host-port") }}`  to match with the `if` section above. Right ?
```
        {{- if .Values.agent.extraEnv }}
          {{- toYaml .Values.agent.extraEnv | nindent 10 }}
        {{- end }}
        {{- if not (hasKey .Values.agent.cmdlineParams "reporter.grpc.host-port") }}
          - name: REPORTER_GRPC_HOST_PORT
            value: {{ include "jaeger.collector.name" . }}:{{ .Values.collector.service.grpc.port }}
        {{- end }
```

charts/jaeger/templates/collector-deploy.yaml
```diff
@@ -66,6 +66,7 @@ spec:
           - name: {{ $key | quote }}
             value: {{ $value | quote }}
           {{- end }}
+          {{- toYaml .Values.storage.kafka.extraEnv | nindent 10 }}
```
mehta-ankit, line 69: I think we need the if check here `{{- if .Values.storage.kafka.extraEnv }}`

charts/jaeger/templates/ingester-deploy.yaml
```diff
@@ -57,6 +57,7 @@ spec:
           {{- end }}
           {{- include "storage.cmdArgs" . | nindent 10 }}
         env:
+          {{- toYaml .Values.ingester.extraEnv | nindent 10 }}
```
mehta-ankit, line 60: same as the comment above.

charts/jaeger/templates/agent-ds.yaml
naseemkullah: done!

charts/jaeger/templates/ingester-deploy.yaml
naseemkullah: good catch, thanks!

charts/jaeger/templates/collector-deploy.yaml
naseemkullah: good catch, thanks!

charts/jaeger/values.yaml
```diff
@@ -431,7 +440,7 @@ spark:
   podLabels: {}
 
 esIndexCleaner:
-  enabled: false
+  enabled: true
```
mehta-ankit, line 443: Just curious, why are we making this true by default ?
Also, if we are doing  so, are we adding it to notes or README ? (for some people this might be a breaking change if they were not expecting to loose data from es-index-cleaner cron job (as it was false before)

naseemkullah: This is just me being sloppy, good catch!

Jasstkn:
Sorry, I was away from Github for a few days. LGTM too.

Issue #126, opened 2020-06-19
[jaeger-operator] jaeger-operator-metrics service shouldn't be created by the chart


jpkrohling:
The `jaeger-operator-metrics` service seems to be [created by the Helm Chart](https://github.com/jaegertracing/helm-charts/blob/fa4dac8d7705fad8a02733aa43019e14dab0b547/charts/jaeger-operator/templates/service.yaml#L4), but it's actually created by the jaeger-operator [during its bootstrap](https://github.com/jaegertracing/jaeger-operator/blob/ec535126223c87c7358ef1dfb6c4af588d4e07b3/pkg/cmd/start/bootstrap.go#L355-L377). The service from the chart doesn't seem to be creating a properly configured service, as seen in a [recent thread](https://groups.google.com/d/msg/jaeger-tracing/CbiPL10p7Qo/r4F1Cx9RBwAJ) in the Jaeger mailing list.

naseemkullah:
cc @cpanato @batazor

scott-dang91:
any update?

PR #127, opened 2020-06-19
[jaeger-operator] fix jaeger-operator chart bug

zackzhangkai:
This is a bug when using chart to deploy jaeger-operator.

The selector of service is not match the label of pods.

And after delete the jaeger-operator-metrics service and jaeger-operator pod, only the pod can recreate but the service cannot.


zackzhangkai:
refered issure: https://github.com/jaegertracing/jaeger-operator/issues/1076

naseemkullah:
cc @cpanato @batazor

cpanato:
@zackzhangkai please bump the chart version

zackzhangkai:
jaeger-operator-1.17.1 and jaeger-operator-1.18.0  have the problem.

naseemkullah:
I don't use jaeger operator but I am quite familiar with helm charts, my two cents is to address the problem here: https://github.com/jaegertracing/helm-charts/blob/fa4dac8d7705fad8a02733aa43019e14dab0b547/charts/jaeger-operator/templates/_helpers.tpl#L45-L48

naseemkullah:
I don't see a reason to stray away from what `helm create` does:

https://github.com/helm/helm/blob/561cc4280830af4a40f2e60c5f17fb3ee69b4e39/pkg/chartutil/create.go#L421-L439

naseemkullah:
Maybe the fix could be addressed in https://github.com/jaegertracing/helm-charts/blob/fa4dac8d7705fad8a02733aa43019e14dab0b547/charts/jaeger-operator/templates/_helpers.tpl#L45-L48 by doing what `helm create` does: 

https://github.com/helm/helm/blob/561cc4280830af4a40f2e60c5f17fb3ee69b4e39/pkg/chartutil/create.go#L421-L439

zackzhangkai:
already fix it as your suggesstion

naseemkullah:
> already fix it as your suggesstion

I don't think that will work. The selector labels are  `app.kubernetes.io/name` and
    `app.kubernetes.io/instance` so the common labels should include at least that, I would suggest to please download latest version of `helm` cli and run `helm create foo` and have a look at the generated helpers, service and deployment templates, then do the same thing here. 😄

naseemkullah:
i.e.

```yaml
{{/*
Common labels
*/}}
{{- define "jaeger-operator.labels" -}}
helm.sh/chart: {{ include "jaeger-operator.chart" . }}
{{ include "jaeger-operator.selectorLabels" . }}
{{- if .Chart.AppVersion }}
app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
{{- end }}
app.kubernetes.io/managed-by: {{ .Release.Service }}
{{- end }}

{{/*
Selector labels
*/}}
{{- define "jaeger-operator.selectorLabels" -}}
app.kubernetes.io/name: {{ include "jaeger-operator.name" . }}
app.kubernetes.io/instance: {{ .Release.Name }}
{{- end }}
```

zackzhangkai:
The service should not be created by this chart  as this [issue](https://github.com/jaegertracing/helm-charts/issues/126), so I remove it.

naseemkullah:
So then `app.kubernetes.io/name` does not need to change, right? Just remove the service template as you've done, but leave the label as is.

charts/jaeger-operator/templates/NOTES.txt
```diff
@@ -2,7 +2,7 @@ jaeger-operator is installed.
 
 
 Check the jaeger-operator logs
-  export POD=$(kubectl get pods-l app.kubernetes.io/instance={{ .Release.Name }} -lapp.kubernetes.io/name=jaeger-operator --namespace {{ .Release.Namespace }} --output name)
+  export POD=$(kubectl get pods-l app.kubernetes.io/instance={{ .Release.Name }} -lname=jaeger-operator --namespace {{ .Release.Namespace }} --output name)
```
naseemkullah, line 5: Please revert.

charts/jaeger-operator/templates/_helpers.tpl
```diff
@@ -44,5 +44,5 @@ Create chart name and version as used by the chart label.
 
 {{/* Generate basic labels */}}
 {{- define "jaeger-operator.labels" }}
-app.kubernetes.io/name: {{ include "jaeger-operator.name" . }}
```
naseemkullah, line 47: Please revert.

zackzhangkai:
This is actually what is the problem. The selector of the service `jaeger-operator-metrics` created by itself , is always `name=jaeger-operator`. 

just as follows :
```
 # k get po jaeger-operator-5dbf4647d5-tkvhx --show-labels
NAME                               READY   STATUS    RESTARTS   AGE   LABELS
jaeger-operator-5dbf4647d5-tkvhx   1/1     Running   0          17s   app.kubernetes.io/name=jaeger-operator,pod-template-hash=5dbf4647d5

# k describe svc jaeger-operator-metrics
Name:              jaeger-operator-metrics
Namespace:         istio-system
Labels:            name=jaeger-operator
Annotations:       <none>
Selector:          name=jaeger-operator
Type:              ClusterIP
IP:                10.233.30.93
Port:              http-metrics  8383/TCP
TargetPort:        8383/TCP
Endpoints:         <none>
Port:              cr-metrics  8686/TCP
TargetPort:        8686/TCP
Endpoints:         <none>
Session Affinity:  None
```

naseemkullah:
> This is actually what is the problem. The selector of the service `jaeger-operator-metrics` created by itself , is always `name=jaeger-operator`.
> 
> just as follows :
> 
> ```
>  # k get po jaeger-operator-5dbf4647d5-tkvhx --show-labels
> NAME                               READY   STATUS    RESTARTS   AGE   LABELS
> jaeger-operator-5dbf4647d5-tkvhx   1/1     Running   0          17s   app.kubernetes.io/name=jaeger-operator,pod-template-hash=5dbf4647d5
> 
> # k describe svc jaeger-operator-metrics
> Name:              jaeger-operator-metrics
> Namespace:         istio-system
> Labels:            name=jaeger-operator
> Annotations:       <none>
> Selector:          name=jaeger-operator
> Type:              ClusterIP
> IP:                10.233.30.93
> Port:              http-metrics  8383/TCP
> TargetPort:        8383/TCP
> Endpoints:         <none>
> Port:              cr-metrics  8686/TCP
> TargetPort:        8686/TCP
> Endpoints:         <none>
> Session Affinity:  None
> ```
> 
> So it is necessary to change the labels of the pods.

Oh! I see ok. I clearly do not know enough about jaeger operator (today, my org uses Jaeger directly via https://github.com/jaegertracing/helm-charts/tree/master/charts/jaeger). I will let @cpanato and @batazor review this one just like all other operator PRs.

zackzhangkai:
Hi, what's going on?

naseemkullah:
@zackzhangkai the PR is awaiting review from jaeger operator chart maintainers.

appanycd:
@batazor @cpanato Can you please review this fix?

zackzhangkai:
Hi, how about this? It has been for a bit long time .  What's going on?  Could you please verify this ? Because our system also has this problem, and we want to be the same with the community . Thank you.

charts/jaeger-operator/templates/service.yaml
```diff
@@ -1,16 +0,0 @@
-apiVersion: v1
```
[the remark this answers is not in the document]
zackzhangkai: this service is created by itself, whenever the backend pod is created, should not install this service mannually.

PR #133, opened 2020-07-13
[jaeger] added image pull secrets option

fahlke:


naseemkullah:
Thank you @fahlke, WDYT about using `with` and `toYaml` for this like in `helm create` generated output: https://github.com/helm/helm/blob/7dd42d7c061436c7d0f1fa5d5246616e8bfbb8b9/pkg/chartutil/create.go#L264-L267 ?

fahlke:
@naseemkullah you're absolutely right. Would improve it by a lot! :) I'll adjust that later today.

naseemkullah:
Great, thanks @fahlke!

charts/jaeger/templates/agent-sa.yaml
```diff
@@ -6,4 +6,8 @@ metadata:
   labels:
     {{- include "jaeger.labels" . | nindent 4 }}
     app.kubernetes.io/component: agent
+{{- with .Values.agent.imagePullSecrets }}
+imagePullSecrets:
+  {{- toYaml . | nindent 0 }}
+{{- end }}
```
naseemkullah, line 12: No need for this in a service account.

charts/jaeger/templates/cassandra-schema-sa.yaml
```diff
@@ -6,4 +6,8 @@ metadata:
   labels:
     {{- include "jaeger.labels" . | nindent 4 }}
     app.kubernetes.io/component: cassandra-schema
+{{- with .Values.schema.imagePullSecrets }}
+imagePullSecrets:
+  {{- toYaml . | nindent 0 }}
+{{- end }}
```
naseemkullah, line 12: No need for this in a service account.

fahlke:
If someone is opting out of the creation of the ServiceAccounts for whatever reason, the imagePullSecrets would have no effect in the current implementation.

I could also implement a switch for checking if SA is created in each Job, Daemonset, etc or not and then add the `with` and `toYaml` there as well.

@naseemkullah What do you think?

naseemkullah:
> If someone is opting out of the creation of the ServiceAccounts for whatever reason, the imagePullSecrets would have no effect in the current implementation.
> 
> I could also implement a switch for checking if SA is created in each Job, Daemonset, etc or not and then add the `with` and `toYaml` there as well.
> 
> @naseemkullah What do you think?

Sorry my mistake. I'll review again with now having learned https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-image-pull-secret-to-service-account

naseemkullah:
retracting comments...

charts/jaeger/templates/agent-sa.yaml
```diff
@@ -6,4 +6,8 @@ metadata:
   labels:
     {{- include "jaeger.labels" . | nindent 4 }}
     app.kubernetes.io/component: agent
+{{- with .Values.agent.imagePullSecrets }}
+imagePullSecrets:
+  {{- toYaml . | nindent 0 }}
```
naseemkullah, line 11:
```suggestion
  {{- toYaml . | nindent 2 }}
```

charts/jaeger/templates/cassandra-schema-sa.yaml
```diff
@@ -6,4 +6,8 @@ metadata:
   labels:
     {{- include "jaeger.labels" . | nindent 4 }}
     app.kubernetes.io/component: cassandra-schema
+{{- with .Values.schema.imagePullSecrets }}
+imagePullSecrets:
+  {{- toYaml . | nindent 0 }}
```
naseemkullah, line 11:
```suggestion
  {{- toYaml . | nindent 2 }}
```

charts/jaeger/templates/collector-sa.yaml
```diff
@@ -6,4 +6,8 @@ metadata:
   labels:
     {{- include "jaeger.labels" . | nindent 4 }}
     app.kubernetes.io/component: collector
+{{- with .Values.collector.imagePullSecrets }}
+imagePullSecrets:
+  {{- toYaml . | nindent 0 }}
```
naseemkullah, line 11:
```suggestion
  {{- toYaml . | nindent 2 }}
```

charts/jaeger/templates/es-index-cleaner-sa.yaml
```diff
@@ -6,4 +6,8 @@ metadata:
   labels:
     {{- include "jaeger.labels" . | nindent 4 }}
     app.kubernetes.io/component: es-index-cleaner
+{{- with .Values.esIndexCleaner.imagePullSecrets }}
+imagePullSecrets:
+  {{- toYaml . | nindent 0 }}
```
naseemkullah, line 11:
```suggestion
  {{- toYaml . | nindent 2 }}
```

charts/jaeger/templates/hotrod-sa.yaml
```diff
@@ -6,4 +6,8 @@ metadata:
   labels:
     {{- include "jaeger.labels" . | nindent 4 }}
     app.kubernetes.io/component: hotrod
+{{- with .Values.hotrod.image.pullSecrets }}
+imagePullSecrets:
+  {{- toYaml . | nindent 0 }}
```
naseemkullah, line 11:
```suggestion
  {{- toYaml . | nindent 2 }}
```

charts/jaeger/templates/ingester-sa.yaml
```diff
@@ -6,4 +6,8 @@ metadata:
   labels:
     {{- include "jaeger.labels" . | nindent 4 }}
     app.kubernetes.io/component: ingester
+{{- with .Values.ingester.imagePullSecrets }}
+imagePullSecrets:
+  {{- toYaml . | nindent 0 }}
```
naseemkullah, line 11:
```suggestion
  {{- toYaml . | nindent 2 }}
```

charts/jaeger/templates/query-sa.yaml
```diff
@@ -6,4 +6,8 @@ metadata:
   labels:
     {{- include "jaeger.labels" . | nindent 4 }}
     app.kubernetes.io/component: query
+{{- with .Values.query.imagePullSecrets }}
+imagePullSecrets:
+  {{- toYaml . | nindent 0 }}
```
naseemkullah, line 11:
```suggestion
  {{- toYaml . | nindent 2 }}
```

charts/jaeger/templates/spark-sa.yaml
```diff
@@ -6,4 +6,8 @@ metadata:
   labels:
     {{- include "jaeger.labels" . | nindent 4 }}
     app.kubernetes.io/component: spark
+{{- with .Values.spark.imagePullSecrets }}
+imagePullSecrets:
+  {{- toYaml . | nindent 0 }}
```
naseemkullah, line 11:
```suggestion
  {{- toYaml . | nindent 2 }}
```

naseemkullah:
> If someone is opting out of the creation of the ServiceAccounts for whatever reason, the imagePullSecrets would have no effect in the current implementation.
> 
> I could also implement a switch for checking if SA is created in each Job, Daemonset, etc or not and then add the `with` and `toYaml` there as well.
> 
> @naseemkullah What do you think?

Are there any issues with, in all cases, adding the imagePullSecrets in the deployments, instead of conditionally in sa or deployment?

charts/jaeger/templates/agent-sa.yaml
fahlke: Sure. I was sticking to what was rendered out in the other templates. I'll update that to `nindent 2`

Same applies for the other changes as suggested by you, as well.

naseemkullah: 🙏 

Thanks! I'll try to clean up the lists to be indented in a future PR to avoid this confusion.

fahlke:
> Are there any issues with, in all cases, adding the imagePullSecrets in the deployments, instead of conditionally in sa or deployment?

I need to think about it for a bit. I guess this is more or less a "philosophical" question. It would not break anything if we use either ServiceAccount or Deployment/Job/Daemonset. It depends on what the end-user likes.

I was more or less following the PR for the Jaeger-Operator where the imagePullSecrets were added:
- PR: https://github.com/jaegertracing/helm-charts/pull/44
- Commit: https://github.com/jaegertracing/helm-charts/pull/44/commits/95d4a92c5927b666128f9b3489473c18151772b1

There it's also conditional in the Deployment for example.

naseemkullah:
Ah I see thanks. Thanks for looking into consistency, but fwiw the two charts are independently managed.

The advantage to having it in the deployment as per https://github.com/helm/helm/blob/f90e5ba316a71f38f160fed55bb3599093410436/pkg/chartutil/create.go#L263-L267 is that it is no longer dependent on the SA.

We'd have to find an advantage to it being in the SA to opt for it imho.

fahlke:
I cannot really find any huge advantages for the `imagePullSecrets` being in the SA's besides referencing to one of my favorite books _"Kubernetes Best Practices: Blueprints for Building Successful Applications on Kubernetes"_ [1] and that the _Kubernetes guides_ [2] also refer to that option.

- [1] Book: https://books.google.de/books?id=Cju-DwAAQBAJ&lpg=PA62&ots=AsSLtE67F2&dq=kubernetes+best+practices+imagePullSecrets&hl=de&pg=PA62&redir_esc=y#v=onepage&q=kubernetes%20best%20practices%20imagePullSecrets&f=false
- [2] K8s guides: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-imagepullsecrets-to-a-service-account

Besides that, keeping it consistent with the independently managed jaeger-operator, was my main idea in the first place.

I'm in for all options.

1. have the `imagePullSecrets` only in the Pod specs (in the Deployments, Jobs, Daemonsets, ...)
2. make it conditionally set by Helm for users who are using SA's or opt-out (by default the SA's are created anyways)
3. _anything else..._

naseemkullah:
Thanks for the links!

FWIW I've found this issue, it would perhaps be applicable in the context of a fresh chart installation https://github.com/kubernetes/kubernetes/issues/88456

naseemkullah:
If we were not already confused enough: https://kubernetes.io/docs/concepts/containers/images/#specifying-imagepullsecrets-on-a-pod

According to k8s docs, it is the recommended approach (to declare in pod spec)

> Note: This is the recommended approach to run containers based on images in private registries.


Based on this I would opt for 1.

naseemkullah:
But it's strange that the book recommends SA... 🤷‍♂️

fahlke:
Awesome! I did not read that. I guess we have a winner! :)

I would suggest that we follow the recommended way from the K8s-docs and put the `imagePullSecrets` configuration into the Pod-Spec's.

I'll adjust revert my changes in the SA's and add the `imagePullSecrets` config to all Pod-Spec's.

fahlke:
@naseemkullah I guess the one open request is obsolete now, as there are no changes in the SA's anymore.

Changed every `imagePullSecrets` to Pod specs as we discussed yesterday.

naseemkullah:
Thank you @fahlke! One last thing, this repo requires that commits be [GPG signed](https://docs.github.com/en/github/authenticating-to-github/about-commit-signature-verification). Could you please GPG sign your commit?

fahlke:
Sure, absolutely. I did not read that the GPG signature is mandatory. Is that stated somewhere "hidden"? I cannot find it here: https://github.com/jaegertracing/helm-charts/blob/master/CONTRIBUTING.md

Anyways. Signed-off and GPG signed now. :)

Next time, I'll more closely read the Contribution Guidelines.

naseemkullah:
Great, thanks @fahlke!

PR #136, opened 2020-07-27
[jaeger] Have an ability to make collector svc headless

spacentropy:
Hi. 
gRPC client balancing requires headless services in order to work. More info here - <https://kubernetes.io/blog/2018/11/07/grpc-load-balancing-on-kubernetes-without-tears/>
This change allows to set collector service type to headless. We use it in production without problems.

### DIFF
--- a/charts/jaeger/templates/collector-svc.yaml
+++ b/charts/jaeger/templates/collector-svc.yaml
@@ -57,9 +57,9 @@ spec:
   selector:
     {{- include "jaeger.selectorLabels" . | nindent 4 }}
     app.kubernetes.io/component: collector
+  {{- if and (eq .Values.collector.service.type "ClusterIP") .Values.collector.service.clusterIP }}
+  clusterIP: {{ .Values.collector.service.clusterIP }}
+  {{- end }}
   type: {{ .Values.collector.service.type }}
-{{- if and (eq .Values.collector.service.type "LoadBalancer") .Values.collector.service.loadBalancerIP }}
-  loadBalancerIP: {{ .Values.collector.service.loadBalancerIP }}
-{{- end -}}
 {{- template "loadBalancerSourceRanges" .Values.collector }}
 {{- end -}}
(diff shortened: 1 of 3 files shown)


charts/jaeger/templates/collector-svc.yaml
```diff
@@ -41,6 +41,7 @@ spec:
   selector:
     {{- include "jaeger.selectorLabels" . | nindent 4 }}
     app.kubernetes.io/component: collector
+  clusterIP: {{ .Values.collector.service.clusterIP }}
```
naseemkullah, line 44:
```suggestion
  {{- if and (eq .Values.collector.service.type "ClusterIP") .Values.collector.service.clusterIP }}
  clusterIP: {{ .Values.collector.service.clusterIP }}
  {{- end }}
```

As to avoid `clusterIP: ` rendering with no value. Also have seen some bad behaviour when that is not handled: https://github.com/helm/helm/issues/7956

pavelnikolov: We could also check if `Values.clusterIP` is not `""`

pavelnikolov:
@spacentropy are you planning to add the `if` statement that @naseemkullah has suggested above?

spacentropy:
Hi. Sorry for the Loooooong delay. I've rebased the branch and applied the suggested changes.

Thanks!

spacentropy:
Hi! Let's return to this. I've made the requested changes. Can we proceed to merging this? @pavelnikolov

pavelnikolov (approved):

spacentropy:
@pavelnikolov Thanks for the approval! Can you please merge? I've rebased & signed commits.

indreek:
Why did you remove loadBalancerIP field? If I plan to run only one instance of collector, i don't need Linkerd in front. It would be fully customizable if you wish to use k8s lb.

```
{{- if and (eq .Values.collector.service.type "LoadBalancer") .Values.collector.service.loadBalancerIP }}
  loadBalancerIP: {{ .Values.collector.service.loadBalancerIP }}
{{- end -}}
```

PR #138, opened 2020-07-29
[jaeger-operator] Fix whitespace typo in post-install notes

llamasoft:
Just a minor typographic fix.  The command as originally written doesn't work because `get pods-l` isn't a valid kubectl command.


naseemkullah:
Thanks @llamasoft, the chart version will need to be bumped and the DCO will need to be signed to get this in.

PR #140, opened 2020-08-11
Allow overriding the Cassandra schema job security context

chgl:
Similar to the existing deployment objects, this adds support for overriding the Cassandra schema job's securityContext.


naseemkullah:
Thanks @chgl! This repo requires that commits be [GPG signed](https://docs.github.com/en/github/authenticating-to-github/about-commit-signature-verification). Could you please address that? Then we can merge.

chgl:
> Thanks @chgl! This repo requires that commits be [GPG signed](https://docs.github.com/en/github/authenticating-to-github/about-commit-signature-verification). Could you please address that? Then we can merge.

Great! Signing seems to have worked.

PR #147, opened 2020-09-08
add CODEOWNERS and remove OWNERS files

naseemkullah:
cc @scottrigby


.github/workflows/CODEOWNERS
```diff
@@ -0,0 +1,4 @@
+# https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
```
scottrigby, line 1: Only actions should BFS in the workflows directory. CODEOWNERS can be in one of a few locations including the .github dir (see the doc for details)

scottrigby (dismissed):
Otherwise looks good.

Question: what about files not in one of the charts directories. Do you want to assign `*` to individual repo admin users, or perhaps a team?

.github/workflows/CODEOWNERS
naseemkullah: oops yeah, thanks

Issue #149, opened 2020-09-10
[jaeger] Kafka pod keep restarting because of SIGTERM signal


mehta-ankit:
Using Latest version of the chart `0.36.0`, and installing it with ingester `enabled`  and with new kafka cluster.

Zookeeper pods start up normally, but kafka pod keeps shutting down because of a SIGTERM signal as seen in the logs.

Before the SIGTERM signal:
![image](https://user-images.githubusercontent.com/22157398/92812737-d522fe00-f38d-11ea-84e2-1a2e0749c65f.png)

After the SIGTERM signal
![image](https://user-images.githubusercontent.com/22157398/92812717-cc322c80-f38d-11ea-8f75-a9d742abc34a.png)

Based on the README, there shouldn't be anything else needed to run this setup with this chart.

Any ideas why Kafka is failing to come up.

mehta-ankit:
I had to tweak liveness and readiness probe values, and kafka pods (3 in my case because of replica value being 3) were able to come up and go into ready state.

mehta-ankit:
@naseemkullah have you tried to use the kafka chart that our helm chart points to ?
I am still concerned because even after the pods were up, they kept failing the liveness probe.
I checked kafka chart uses `jps` as a liveness probe check. I tried running jps by doing exec on the pod, but it returns nothing, even when a java process is up and running.

I checked bitnami's Kafka chart does not use jps but instead uses tcp-socket: https://github.com/bitnami/charts/blob/master/bitnami/kafka/templates/statefulset.yaml#L298

Also i found this bug, https://bugs.openjdk.java.net/browse/JDK-8193710.
Which i think can maybe mess up the liveness probe.

SO i wanted to check if others have used this kafka chart and can confirm it works fine.

naseemkullah:
Hey @mehta-ankit, thanks for raising this. I have never personally installed the Kafka component. Seeing that the incubator repo is deprecated we should try another Kafka chart as a dependency. Maybe the bitnami one would be better?

naseemkullah:
I though confluence charts would be a good idea initially https://github.com/confluentinc/cp-helm-charts/issues/374 but no response

mehta-ankit:
> Hey @mehta-ankit, thanks for raising this. I have never personally installed the Kafka component. Seeing that the incubator repo is deprecated we should try another Kafka chart as a dependency. Maybe the bitnami one would be better?

I can give bitnami chart a try and update the issue. 

[...]

PR #150, opened 2020-09-15
Use bitnami chart for kafka

mehta-ankit:

Resolves https://github.com/jaegertracing/helm-charts/issues/149

Incubator kafka chart will be obsolete soon as per these instructions: https://github.com/helm/charts#deprecation-timeline
Also liveness probe was failing for the incubator kafka chart.

Bitnami provides stable kafka chart that Jaeger's helm chart can use.

### DIFF
--- a/charts/jaeger/values.yaml
+++ b/charts/jaeger/values.yaml
@@ -90,11 +90,13 @@ cassandra:
 
 # Begin: Override values on the Kafka subchart to customize for Jaeger
 kafka:
-  replicas: 1
-  configurationOverrides:
-    "auto.create.topics.enable": true
+  replicaCount: 1
+  autoCreateTopicsEnable: true
   zookeeper:
     replicaCount: 1
+    serviceAccount:
+      create: true
+
 # End: Override values on the Kafka subchart to customize for Jaeger
 
 # Begin: Default values for the various components of Jaeger
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -33,6 +33,7 @@ jobs:
           helm repo add stable https://kubernetes-charts.storage.googleapis.com/
           helm repo add incubator https://kubernetes-charts-incubator.storage.googleapis.com/
           helm repo add elastic https://helm.elastic.co
+          helm repo add bitnami https://charts.bitnami.com/bitnami
 
       - name: Run chart-releaser
         uses: helm/chart-releaser-action@v1.0.0-rc.2
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 # Helm chart automated files
 /charts/*/charts
 /charts/*/requirements.lock
+/charts/*/*.tgz
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.19.2
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
-version: 0.36.0
+version: 0.37.0
 keywords:
   - jaeger
   - opentracing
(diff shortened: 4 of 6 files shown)


charts/jaeger/values.yaml
```diff
@@ -90,11 +90,8 @@ cassandra:
 
 # Begin: Override values on the Kafka subchart to customize for Jaeger
 kafka:
-  replicas: 1
-  configurationOverrides:
-    "auto.create.topics.enable": true
-  zookeeper:
-    replicaCount: 1
+  autoCreateTopicsEnable: true
+
```
mehta-ankit, line 94: by default, kafka and zookeeper replica values are 1 for bitnami chart.

mehta-ankit:
these are the values i used to test it on my cluster.
```
# Values for bitnami kafka chart: https://github.com/bitnami/charts/blob/master/bitnami/kafka/values.yaml
  kafka:
    replicaCount: 3
    podSecurityContext:
      # TIL: https://github.com/bitnami/bitnami-docker-kafka/blob/master/2/debian-10/Dockerfile#L30
      # bitnami builds its images with non-root user, but if you don't use the same user, you don't
      # have permissions to read/write on those folders on the mounted volume.
      runAsUser: 1001
      fsGroup: 1001
    resources:
      limits:
        cpu: 500m
        memory: 2Gi
      requests:
        cpu: 250m
        memory: 1Gi
    metrics:
      kafka:
        enabled: true
      jmx:
        enabled: true
    serviceAccount:
      create: true
    # Values for zookepr bitnasmi chart: https://github.com/bitnami/charts/blob/master/bitnami/zookeeper/values.yaml
    zookeeper:
      securityContext:
        enabled: true
        fsGroup: 1001
        runAsUser: 1001
      replicaCount: 3
      metrics:
        enabled: true
      serviceAccount:
        create: true
```

@naseemkullah do you think, if any of those should be default in jaeger's values.yaml file ?
The only one i could think of is `serviceAccount.create: true`. Let me know what you think.

naseemkullah:
> these are the values i used to test it on my cluster.
> 
> ```
>   kafka:
>     replicaCount: 3
>     podSecurityContext:
>       runAsUser: 1001
>       fsGroup: 1001
>     resources:
>       limits:
>         cpu: 500m
>         memory: 2Gi
>       requests:
>         cpu: 250m
>         memory: 1Gi
>     metrics:
>       kafka:
>         enabled: true
>       jmx:
>         enabled: true
>     serviceAccount:
>       create: true
>     zookeeper:
>       securityContext:
>         enabled: true
>         fsGroup: 1001
>         runAsUser: 1001
>       replicaCount: 3
>       metrics:
>         enabled: true
>       serviceAccount:
>         create: true
> ```
> 
> @naseemkullah do you think, if any of those should be default in jaeger's values.yaml file ?
> The only one i could think of is `serviceAccount.create: true`. Let me know what you think.

Hard to say, I'd add the minimal config required for it to work out of the box and not fall over. `serviceAccount: true` does seem like a good idea for `zoopkeeper` (though it appears to be already enabled for `kafka` according to the chart).

Things like a sane securityContext would ideally be configured in the subcharts' default values. My two cents. As I do not plan to use this feature I can't really say much more.

@arpitjindal97 seeing that you originally added this feature please take a look once it is ready for review if you can!

mehta-ankit:
Ready for review!

naseemkullah (approved):
lgtm, but will wait to see if @arpitjindal97 can take a look before merging

charts/jaeger/values.yaml
arpitjindal97: One change needed here. Although default replica set will be 1 for `kafka` and `zookeeper` but It would be better to set it explicitly. When any user wants to change this value, he/she wouldn't have to go to Bitnami's Documentation and find correct `key`

mehta-ankit: Sure. Although one could make that same argument for few other key's 😄 . I ll push that change up.

naseemkullah (approved):

naseemkullah:
Thanks @mehta-ankit!

Issue #153, opened 2020-09-23
[jaeger] extraEnv variables need to be passed as strings


user-name-is-taken:
This section

```
storage:
  cassandra:
    extraEnv: []
      # - name: CASSANDRA_SERVERS
      #   value: cassandra
      # - name: CASSANDRA_PORT
      #   value: 9042
      # - name: CASSANDRA_KEYSPACE
      #   value: jaeger_v1_test
      # - name: CASSANDRA_TLS_ENABLED
      #   value: false
```

Should be

```
storage:
  cassandra:
    extraEnv: []
      # - name: CASSANDRA_SERVERS
      #   value: cassandra
      # - name: CASSANDRA_PORT
      #   value: "9042"
      # - name: CASSANDRA_KEYSPACE
      #   value: jaeger_v1_test
      # - name: CASSANDRA_TLS_ENABLED
      #   value: "false"
```

because env vars have to be strings as discussed [here](https://github.com/helm/helm/issues/4775#issuecomment-431171187)

naseemkullah:
Thanks for spotting this @user-name-is-taken , would you be willing to open a PR to fix this?

user-name-is-taken:
[resolved in this MR](https://github.com/jaegertracing/helm-charts/pull/154#event-3799684157)

PR #154, opened 2020-09-23
Resolving issue #153

user-name-is-taken:
#153

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.19.2
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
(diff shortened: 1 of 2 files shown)


naseemkullah:
Thanks @user-name-is-taken I just created a PR template, but you beat me to it with your commit :) 

As per https://github.com/jaegertracing/helm-charts/blob/master/.github/PULL_REQUEST_TEMPLATE.md#checklist the DCO must be signed, commits must be GPG signed and chart version must be bumped (though we can technically let the pipeline fail for this one since its just a comment change). 

Cheers and thanks for contributing!

naseemkullah:
Almost there, just need the GPG signature on the commit(s): https://docs.github.com/en/github/authenticating-to-github/about-commit-signature-verification

PR #160, opened 2020-09-30
[jaeger] Add elasticsearch rollover jobs

n4nn31355:
Signed-off-by: Vyacheslav Artemiev <v.artemiev@semrush.com>

#### What this PR does

Adds jobs to use the ES rollover feature https://www.jaegertracing.io/docs/1.19/deployment/#elasticsearch-rollover

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/templates/_helpers.tpl
+++ b/charts/jaeger/templates/_helpers.tpl
@@ -85,6 +85,28 @@ Create the name of the esIndexCleaner service account to use
 {{- end -}}
 
 {{/*
+Create the name of the esRollover service account to use
+*/}}
+{{- define "jaeger.esRollover.serviceAccountName" -}}
+{{- if .Values.esRollover.serviceAccount.create -}}
+  {{ default (printf "%s-es-rollover" (include "jaeger.fullname" .)) .Values.esRollover.serviceAccount.name }}
+{{- else -}}
+  {{ default "default" .Values.esRollover.serviceAccount.name }}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Create the name of the esLookback service account to use
+*/}}
+{{- define "jaeger.esLookback.serviceAccountName" -}}
+{{- if .Values.esLookback.serviceAccount.create -}}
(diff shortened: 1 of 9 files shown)


charts/jaeger/templates/_helpers.tpl
```diff
@@ -84,6 +84,28 @@ Create the name of the esIndexCleaner service account to use
...
+{{- define "jaeger.esRollover.serviceAccountName" -}}
+{{- if .Values.esRollover.serviceAccount.create -}}
+  {{ default (printf "%s-es-rollover" (include "jaeger.fullname" .)) .Values.esRollover.serviceAccount.name }}
+{{- else -}}
+  {{ default "default" .Values.esRollover.serviceAccount.name }}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Create the name of the esRollover service account to use
```
naseemkullah, line 99:
```suggestion
Create the name of the esLookback service account to use
```

charts/jaeger/values.yaml
```diff
@@ -487,6 +489,92 @@ esIndexCleaner:
...
+      #   value: "3"
+    annotations: {}
+    podAnnotations: {}
+    podLabels: {}
+    ttlSecondsAfterFinished: 120
+
+esLookback:
+  enabled: false
+  securityContext:
+    runAsUser: 1000
```
naseemkullah, line 542: though it could be used as an override it is redundant as is since `podSecurityContext` value is also 1000, maybe leave empty?

naseemkullah (changes requested):
Wow, thanks! 
Left two small comments

charts/jaeger/values.yaml
n4nn31355: Yep, sounds reasonable to me. Fixed

charts/jaeger/templates/_helpers.tpl
n4nn31355: Thanks. Fixed

naseemkullah (approved):

PR #166, opened 2020-10-08
[jaeger] update readme and add recommendation to explicitly configure storage

naseemkullah:
Signed-off-by: Naseem <naseem@transit.app>

#### What this PR does

Cleans up the readme:

- With the `helm show values` command there is no need to duplicate values in the readme.

- Add note about es rollover requiring a pre existing Elasticsearch (@n4nn31355 can you please confirm?)

- Add note about Jaeger's recommendation for storage

- Add warning that storage should be explicitly configured, as the default settings are subject to change

- Reorganize sections

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.20.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
-version: 0.39.0
+version: 0.39.1
 keywords:
   - jaeger
   - opentracing
--- a/charts/jaeger/README.md
+++ b/charts/jaeger/README.md
@@ -32,9 +32,37 @@ components](https://www.jaegertracing.io/img/architecture-v1.png)
 
 ## Configurable Chart Values
 
+### Jaeger Chart
+
 ```bash
 helm show values jaegertracing/jaeger
 ```
+
+### Dependency Charts
+
+If installing with a dependency such as Cassandra, Elasticsearch and/or Kafka
+their, values can be shown by running:
+
+```bash
+# elasticsearch
+helm repo add elastic https://helm.elastic.co
+helm show values elastic/elasticsearch
+```
+
+```bash
+# cassandra
+helm repo add incubator https://kubernetes-charts-incubator.storage.googleapis.com/
+helm show values incubator/cassandra
+```
+
+```bash
+# bitnami
+helm repo add bitnami
+helm show values bitnami/kafka
+```
+
+Please note, any dependency values must be nested within the key named after the
+chart, i.e. `elasticsearch`, `cassandra` and/or `kafka`.
 
 ## Storage
 
--- a/charts/jaeger/README.md
+++ b/charts/jaeger/README.md
@@ -30,34 +30,52 @@ By default, the chart deploys the following:
 ![Jaeger with Default
 components](https://www.jaegertracing.io/img/architecture-v1.png)
 
-## Configurable Chart Values
+## Configuration
 
-### Jaeger Chart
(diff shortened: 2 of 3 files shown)


n4nn31355:
> Add note about es rollover requiring a pre existing Elasticsearch (@n4nn31355 can you please confirm?)

Oh, yep, that sounds right.

Probably there should be [init containers](https://kubernetes.io/docs/concepts/workloads/pods/init-containers) instead of the hook, but this will require some logic to workaround concurrent runs(when we have more than 1 replica).

naseemkullah:
These changes OK with you @pavelnikolov @dvonthenen @mikelorant?

charts/jaeger/README.md
```diff
@@ -220,182 +236,3 @@ helm install jaeger jaegertracing/jaeger \
   --set storage.kafka.brokers={<BROKER1:PORT>,<BROKER2:PORT>} \
   --set storage.kafka.topic=<TOPIC>
 ```
-
-## Configuration
```
mikelorant, line 224: Was the entire configuration block meant to be deleted? There are a lot more configuration than the Cassandra you removed.

charts/jaeger/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.20.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
-version: 0.39.0
+version: 0.39.1
```
mikelorant, line 5: Considering this will be a breaking change for anyone using Cassandra it would seem likely that the major version needs to be bumped.

charts/jaeger/README.md
naseemkullah: Yes, https://helm.sh/docs/helm/helm_show_values/ will remove the necessity to duplicate these in the README, point taken about sub chart dependencies I will add a note about how to retrieve their values.

charts/jaeger/Chart.yaml
naseemkullah: Today there will be no change in default storage setup, it is an advisory notice that storage should be explicitly configured to be resilient across the change of default storage setup. If this is too confusing I could remove that note altogether?

charts/jaeger/README.md
mikelorant: Is this a standard practise now to remove the configuration from documentation though? Still seems very handy to be able to see what options can be set. Considering the organisation I work for is still on Helm 2, hard to know what the current best practises are so hopefully others can assist here.

naseemkullah: Yes that seems to be the direction charts are moving towards. I take it your org will be moving to helm3 in the near future considering the deprecation timeline?

@scottrigby Could you please chime in?

naseemkullah: Also https://github.com/jaegertracing/helm-charts/pull/166/commits/8cc7aa84ad48f7d32e419b7e8fc7bc7b2bc74355 pushed to address dependency chart values

charts/jaeger/Chart.yaml
pavelnikolov: I am happy with this change. It makes the README independent of any dependency chart changes.

pavelnikolov (approved):

scottrigby (commented):
I commented on the configuration README DRY question.

One other small style note, I'd recommend changing README codeblock linguist option from `bash` to `console` (or `yaml` where appropriate etc). Most of these are really console command prompts not specifically bash scripts.

charts/jaeger/README.md
naseemkullah: Great, thanks @scottrigby. As such I will include the helm 2 equivalent commands. That work for you @mikelorant ?

mikelorant: Works for me.

naseemkullah: Ok great, I've updated the README accordingly.

PR #174, opened 2020-11-28
[jaeger] Add the option to enable ingress for the collector

creack:
#### What this PR does

Add the option to enable ingress for the collector

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/templates/collector-ing.yaml
+++ b/charts/jaeger/templates/collector-ing.yaml
@@ -1,0 +1,30 @@
+{{- if .Values.collector.ingress.enabled -}}
+{{- $servicePort := .Values.collector.service.http.port -}}
+{{- $basePath := .Values.collector.basePath -}}
+apiVersion: networking.k8s.io/v1beta1
+kind: Ingress
+metadata:
+  name: {{ template "jaeger.collector.name" . }}
+  labels:
+    {{- include "jaeger.labels" . | nindent 4 }}
+    app.kubernetes.io/component: collector
+  {{- if .Values.collector.ingress.annotations }}
+  annotations:
+    {{- toYaml .Values.collector.ingress.annotations | nindent 4 }}
+  {{- end }}
+spec:
+  rules:
+    {{- range $host := .Values.collector.ingress.hosts }}
+    - host: {{ $host }}
+      http:
+        paths:
+          - path: {{ $basePath }}
+            backend:
+              serviceName: {{ template "jaeger.collector.name" $ }}
+              servicePort: {{ $servicePort }}
+    {{- end -}}
+  {{- if .Values.collector.ingress.tls }}
+  tls:
+    {{- toYaml .Values.collector.ingress.tls | nindent 4 }}
+  {{- end -}}
+{{- end -}}
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.20.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
-version: 0.39.1
+version: 0.39.2
 keywords:
   - jaeger
   - opentracing
--- a/charts/jaeger/requirements.yaml
+++ b/charts/jaeger/requirements.yaml
@@ -1,6 +1,6 @@
 dependencies:
   - name: cassandra
-    version: ^0.15.0
+    version: 0.15.3
     repository: https://kubernetes-charts-incubator.storage.googleapis.com/
     condition: provisionDataStore.cassandra
   - name: elasticsearch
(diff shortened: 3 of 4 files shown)


creack:
Looks like upstream cassandra 0.15.4 is broken. Pinning to 0.15.3 so the linter can run (and so people can use the chart)

charts/jaeger/templates/collector-ing.yaml
```diff
@@ -0,0 +1,30 @@
+{{- if .Values.collector.ingress.enabled -}}
+{{- $servicePort := .Values.collector.service.http.port -}}
+{{- $basePath := .Values.collector.basePath -}}
+apiVersion: networking.k8s.io/v1beta1
+kind: Ingress
+metadata:
+  name: {{ template "jaeger.collector.name" . }}
+  labels:
+    {{- include "jaeger.labels" . | nindent 4 }}
+    app.kubernetes.io/component: query
```
naseemkullah, line 10:
```suggestion
    app.kubernetes.io/component: collector
```

naseemkullah (changes requested):

naseemkullah (approved):
Thanks!

charts/jaeger/templates/collector-ing.yaml
```diff
@@ -0,0 +1,30 @@
...
+spec:
+  rules:
+    {{- range $host := .Values.collector.ingress.hosts }}
+    - host: {{ $host }}
+      http:
+        paths:
+          - path: {{ $basePath }}
+            backend:
+              serviceName: {{ template "jaeger.collector.name" $ }}
+              servicePort: {{ $servicePort }}
```
mehta-ankit, line 24: @creack @naseemkullah 
I just saw this, and wanted to ask if it was on purpose that servicePort used here in this file is http port (defined here: https://github.com/jaegertracing/helm-charts/blob/master/charts/jaeger/values.yaml#L281)

Asking because before ingress was available in the chart, we used to create our own ingress and we use port 14250 which is for  GRPC (agents talk to collector using GRPC). Only when jaeger clients talk to Collector directly its over HTTP (AFAIK).
So enabling this ingress would not work for agent->collector traffic i feel. Correct if i am wrong though.

creack: It is a copy/paste of the Query ingress (https://github.com/jaegertracing/helm-charts/blob/master/charts/jaeger/templates/query-ing.yaml). 

As this under `spec` -> `rules` -> `http`, I think it makes sense here. I am not very familiar with this, but I suspect to enable grpc, we would need an additional block `spec` -> `rules` -> `grpc`.

mehta-ankit: No even for GRPC traffic rules section stays under GRPC, but there is an added notation needed: 
`nginx.ingress.kubernetes.io/backend-protocol: "GRPC"`
and this i assume would depend on which ingress controller you are using. In our case we are using `nginx ingress`.

The point i am trying to make it, this ingress resource if enabled will not work for Agent -> Collector communication. We should have mentioned it somewhere as IMO that is one of the most common ways people use Jaeger in production (instead of client to collector traffic directly sing HTTP).

naseemkullah: agent -> collector is the recommended approach indeed, we should probably make this port 14250 by default then if thats the most common use case. @creack is that ok with you? If so, @mehta-ankit would you please? That way your org does not need to deploy an ingress separate from this chart

creack: agent->collector is indeed the most common from within the cluster, however, the ingress is meant for the outside. My use-case is to enable the agentless AWS Lambdas, outside the cluster/network to push traces via the ingress.
Maybe we could support both plain http and grpc?

mehta-ankit: @naseemkullah as i mentioned in my previous comment, its not straightforward to handle GRPC traffic by just changing this ingress resource (changing the port number). It comes with its own set of problems.
In our case, we also deploy our own ingress controller with annotation `nginx.ingress.kubernetes.io/backend-protocol: "GRPC"` on the load balancer it spins up. Plus on our AWS load balancer, we add a listener for port 443 and add ALPN policy to allow HTTP/2 traffic.

I am trying to argue that it would be hard to support a default GRPC enabled ingress in this chart in a generic way (at least for now, as its not straight forward. See https://github.com/kubernetes/kubernetes/pull/96861)

My suggestion would be to allow this http ingress (as it just works), but change the boolean condition enabling this ingress to explicitly say its a HTTP ingress for client->collector traffic (not agent->collector GRPC traffic) for now.
```
{{- if .Values.collector.ingress.http.enabled -}}
```
this way in the future we can enable grpc ingress as well, when its simple to do so.

PR #183, opened 2020-12-14
[jaeger] Add concurrencyPolicy to CronJob and maxUnavailable to jaeger-agent daemonset

### LINKED ISSUE #128: [jaeger] Allow concurrencyPolicy to be set from the values.yaml file
**Background ?**
(On Jaeger chart)

Both es-index-cleaner and spark-cronjob do not allow concurrencyPolicy and few other cronJob K8s resources properties to be set from the values.yaml file.
Example: https://github.com/jaegertracing/helm-charts/blob/master/charts/jaeger/templates/es-index-cleaner-cronjob.yaml#L14
This is explicitly set to Forbid which is fine, but in case of [spark-cronjob](https://github.com/jaegertracing/helm-charts/blob/master/charts/jaeger/templates/spark-cronjob.yaml) the property is not set and is [Allow](https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/#concurrency-policy) by default. 
In some scenarios, someone might want to run the spark job more frequently but they would not want multiple jobs to run concurrently (which can cause issues with load on the Elastic-search cluster).

**What ?**

We should make these properties configurable through a helm chart values file and see if there are other properties that make sense to be made configurable as well.

mehta-ankit:

#### What this PR does

- It adds concurrencyPolicy to all cronjob templates, so users can set them as they see fit. Kept the default value "Forbid" for backwards compatibility.
- Added Max Unavailable value to Jaeger agent daemonset (its default 1 and i kept it 1 in the values.yaml file).

- fixes #128 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/templates/agent-ds.yaml
+++ b/charts/jaeger/templates/agent-ds.yaml
@@ -15,6 +15,10 @@ spec:
     matchLabels:
       {{- include "jaeger.selectorLabels" . | nindent 6 }}
       app.kubernetes.io/component: agent
+  {{- with .Values.agent.daemonset.updateStrategy }}
+  updateStrategy:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
   template:
     metadata:
 {{- if .Values.agent.podAnnotations }}
--- a/charts/jaeger/values.yaml
+++ b/charts/jaeger/values.yaml
@@ -196,6 +196,10 @@ agent:
   extraEnv: []
   daemonset:
     useHostPort: false
+    updateStrategy:
+      type: RollingUpdate
+      rollingUpdate:
+        maxUnavailable: 1
   service:
     annotations: {}
     # List of IP ranges that are allowed to access the load balancer (if supported)
@@ -449,6 +453,7 @@ spark:
   schedule: "49 23 * * *"
   successfulJobsHistoryLimit: 5
   failedJobsHistoryLimit: 5
+  concurrencyPolicy: Forbid
   resources: {}
     # limits:
     #   cpu: 500m
@@ -486,6 +491,7 @@ esIndexCleaner:
   schedule: "55 23 * * *"
   successfulJobsHistoryLimit: 3
   failedJobsHistoryLimit: 3
+  concurrencyPolicy: Forbid
   resources: {}
     # limits:
     #   cpu: 500m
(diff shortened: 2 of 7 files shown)


charts/jaeger/templates/agent-ds.yaml
```diff
@@ -28,6 +28,12 @@ spec:
         {{- toYaml .Values.agent.podLabels | nindent 8 }}
 {{- end }}
     spec:
+      updateStrategy:
+        type: {{ .Values.agent.daemonset.updateStrategy }}
+        {{- if eq .Values.agent.daemonset.updateStrategy "RollingUpdate" -}}
+        rollingUpdate:
+          maxUnavailable: {{ .Values.agent.daemonset.maxUnavailable }}
+        {{- end }}
```
naseemkullah, line 36: I think an approach like https://github.com/fluent/helm-charts/blob/e6805d4509868b76717677c7f7eecc4834d49d5a/charts/fluent-bit/templates/daemonset.yaml#L12-L15 will be more flexible and removes unneeded abstraction, WDYT?

naseemkullah (changes requested):
Thanks @mehta-ankit, left a comment.

charts/jaeger/templates/agent-ds.yaml
mehta-ankit: Makes sense @naseemkullah 
Certainly makes it easier. 😄

mehta-ankit: Made the change.

naseemkullah (approved):

charts/jaeger/values.yaml
```diff
@@ -196,6 +196,10 @@ agent:
   extraEnv: []
   daemonset:
     useHostPort: false
+    updateStrategy:
+      type: RollingUpdate
+      rollingUpdate:
+        maxUnavailable: 1
```
naseemkullah, line 202:
```suggestion
    updateStrategy: {}
      # type: RollingUpdate
      # rollingUpdate:
      #   maxUnavailable: 1
```

@mehta-ankit I just merged, would you mind following up with a PR that sets this to empty so that default is implicitly set?

mehta-ankit: https://github.com/jaegertracing/helm-charts/pull/184

PR #205, opened 2021-02-08
[jaeger] Adds ability to set enableServiceLinks to false

koalalorenzo:
#### What this PR does
Attempt to fix: https://github.com/jaegertracing/jaeger/issues/1986 and https://github.com/prometheus/prometheus/issues/7286

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/values.yaml
+++ b/charts/jaeger/values.yaml
@@ -201,7 +201,7 @@ agent:
       # rollingUpdate:
       #   maxUnavailable: 1
   # Set to false if your release is called `jaeger`.
-  # Read More: https://github.com/jaegertracing/jaeger/issues/1986 
+  # Read More: https://github.com/jaegertracing/jaeger/issues/1986
   enableServiceLinks: true
   service:
     annotations: {}
--- a/charts/jaeger/values.yaml
+++ b/charts/jaeger/values.yaml
@@ -200,8 +200,6 @@ agent:
       # type: RollingUpdate
       # rollingUpdate:
       #   maxUnavailable: 1
-  # Set to false if your release is called `jaeger`.
-  # Read More: https://github.com/jaegertracing/jaeger/issues/1986
   enableServiceLinks: true
   service:
     annotations: {}
--- a/charts/jaeger/values.yaml
+++ b/charts/jaeger/values.yaml
@@ -200,6 +200,9 @@ agent:
       # type: RollingUpdate
       # rollingUpdate:
       #   maxUnavailable: 1
+  # Set to false if your release is called `jaeger`.
+  # Read More: https://github.com/jaegertracing/jaeger/issues/1986 
+  enableServiceLinks: true
   service:
     annotations: {}
     # List of IP ranges that are allowed to access the load balancer (if supported)
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.21.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
-version: 0.40.0
+version: 0.40.1
 keywords:
   - jaeger
   - opentracing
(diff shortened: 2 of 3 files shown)


charts/jaeger/values.yaml
```diff
@@ -200,6 +200,9 @@ agent:
       # type: RollingUpdate
       # rollingUpdate:
       #   maxUnavailable: 1
+  # Set to false if your release is called `jaeger`.
+  # Read More: https://github.com/jaegertracing/jaeger/issues/1986
```
naseemkullah, line 204: https://github.com/jaegertracing/jaeger/issues/1986  concerns the query component was worked around via https://github.com/jaegertracing/helm-charts/pull/12, can you plz elaborate on what issue this solves or how to reproduce? Thanks!

naseemkullah (changes requested):

naseemkullah:
So I just started seeing this in my setup as well, i recently moved jaeger into the same namespace as prometheus, and prometheus emits:

```
{"caller":"main.go:560","err":"unable to get tracing config from environment: cannot obtain reporter config from env: cannot parse env var JAEGER_AGENT_PORT=udp://11.2.3.4:5775: strconv.ParseInt: parsing \"udp://1.2.3.4:5775\": invalid syntax","level":"error","msg":"Unable to init tracing","ts":"2021-02-09T16:35:46.791Z"}
````

**Comment by pavelnikolov** [comment]
I've never added the protocol prefix before. Usually use `<ip>:<port>`

**Comment by naseemkullah** [comment]
I think the prometheus server  would have to have `JAEGER_AGENT_PORT` in its env (as done for jaeger query) or `enableServiceLinks` disabled, WDYT @koalalorenzo ?

**Comment by naseemkullah** [comment]
(I've since removed jaeger-agent from our infra in favour of otel-collector agent)

File: charts/jaeger/values.yaml (hunk 1 above)
Comment (koalalorenzo) [reply charts/jaeger/values.yaml:204]: This is supposed to resolve a problem related to exposing jaeger service port via an env variable, causing major issues if the chart name is exactly "jaeger".

**Comment by koalalorenzo** [comment]
@naseemkullah I am not having a good experience with jaeger.

Our **fix** consists into temporary change the `jaeger` name and explore other solutions.
To "fix" it we needed to change `enableServiceLinks` in the manifests of jaeger, as described in [this issue](https://github.com/jaegertracing/jaeger/issues/1986#issuecomment-775183288), so this PR is about that anyway.

**Comment by naseemkullah** [comment]
> @naseemkullah I am not having a good experience with jaeger.
> 
> 
> 
> Our **fix** consists into temporary change the `jaeger` name and explore other solutions to replace entirely jaeger
> 
> 
> 
> To "fix" it we needed to change `enableServiceLinks` in the manifests of jaeger, as described in [this issue](https://github.com/jaegertracing/jaeger/issues/1986#issuecomment-775183288), so this PR is about that anyway.

That issue is about the jaeger query, (I would know I created the issue), I also have a release called jaeger. Please see my comment about Prometheus emitting a log. What issue are you facing? Which component is producing an error?

File: charts/jaeger/values.yaml (hunk 1 above)
Comment (naseemkullah) [inline charts/jaeger/values.yaml:204]:
```suggestion
```

The issue link has been worked around, and for prometheus there are other workarounds such as adding `JAEGER_AGENT_PORT` in prometheus server container's env.

That aside adding the enableServiceLinks bool to the agent and any other components is a welcome addition!

naseemkullah (changes requested):

koalalorenzo:
> It sound like you are facing issues with prometheus erroring out about JAEGER_AGENT_PORT but you would have to confirm.

Yes correct.

naseemkullah (approved):

naseemkullah:
Thanks @koalalorenzo! lgtm, but all checks will need to pass before merging though.

koalalorenzo:
I think that the commit from GitHub adds the issue as it is not signed nor signed with a GPG :(

naseemkullah:
Hi @koalalorenzo GPG seems ok, its the DCO that needs to be signed in this case and conflicts need to be resolved to get this merged.

jonbcampos-alto:
+1

mehta-ankit:
> Hi @koalalorenzo GPG seems ok, its the DCO that needs to be signed in this case and conflicts need to be resolved to get this merged.

@koalalorenzo If you are still looking to push this through. This seems to be the issue:-
image

PR #220, opened 2021-03-17
[jaeger] Allow template strings in kafka brokers value

ctailor2:
#### What this PR does

Pass `.Values.storage.kafka.brokers` through the `tpl` function in both usages to process the value as a template. Allowing template strings here makes it possible for users that provision kafka as part of the jaeger deployment to configure this value in the values yaml file rather than requiring it to be set via a flag. For installs that provision kafka without name overrides, the following template string should generally work here.

```yaml
storage:
  kafka:
    brokers:
      - "{{ .Release.Name }}-kafka"
```

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.22.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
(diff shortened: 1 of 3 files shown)


naseemkullah (approved):
Thanks!

PR #228, opened 2021-03-29
[jaeger-operator] Adding support for NodePort in Jaeger Operator Helm Chart

CSP197:
Signed-off-by: GitHub <noreply@github.com>

#### What this PR does
Adds support for NodePort in the Jaeger Operator

- fixes # https://github.com/jaegertracing/helm-charts/issues/229

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger-operator/Chart.yaml
+++ b/charts/jaeger-operator/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: jaeger-operator Helm chart for Kubernetes
 name: jaeger-operator
-version: 2.19.1
+version: 2.20.0
 appVersion: 1.21.2
 home: https://www.jaegertracing.io/
 icon: https://www.jaegertracing.io/img/jaeger-icon-reverse-color.svg
--- a/charts/jaeger-operator/templates/service.yaml
+++ b/charts/jaeger-operator/templates/service.yaml
@@ -14,4 +14,7 @@ spec:
   selector:
     app.kubernetes.io/name: {{ include "jaeger-operator.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
-  type: ClusterIP
+  type: {{ .Values.service.type }}
+    {{- if and (eq .Values.service.type "NodePort") (.Values.service.nodePort) -}}
+    nodePort: {{ .Values.service.nodePort }}
+  {{- end -}}
--- a/charts/jaeger-operator/templates/service.yaml
+++ b/charts/jaeger-operator/templates/service.yaml
@@ -11,10 +11,10 @@ spec:
     port: 8383
     protocol: TCP
     targetPort: 8383
+{{- if and (eq .Values.service.type "NodePort") (.Values.service.nodePort) }}
+    nodePort: {{ .Values.service.nodePort }}
+{{- end }}
   selector:
     app.kubernetes.io/name: {{ include "jaeger-operator.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
   type: {{ .Values.service.type }}
-    {{- if and (eq .Values.service.type "NodePort") (.Values.service.nodePort) -}}
-    nodePort: {{ .Values.service.nodePort }}
-  {{- end -}}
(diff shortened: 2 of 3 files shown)


charts/jaeger-operator/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: jaeger-operator Helm chart for Kubernetes
 name: jaeger-operator
-version: 2.19.1
+version: 2.19.2
```
cpanato, line 4: small nit: 
```suggestion
version: 2.20.0
```

this is adding a "feature" and not fixing any bug

CSP197: Updated 👍

cpanato (approved):

CSP197:
@batazor Can you review this PR?

charts/jaeger-operator/templates/service.yaml
```diff
@@ -14,4 +14,7 @@ spec:
   selector:
     app.kubernetes.io/name: {{ include "jaeger-operator.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
-  type: ClusterIP
+  type: {{ .Values.service.type }}
+    {{- if and (eq .Values.service.type "NodePort") (.Values.service.nodePort) -}}
+    nodePort: {{ .Values.service.nodePort }}
```
batazor, line 19: Shouldn't this be moved to the port description?

```
ports:
  - name: metrics
    port: 8383
    protocol: TCP
    targetPort: 8383
    {{ if (and (eq .Values.service.type "NodePort") (not (empty .Values.service.nodePort))) }}
    nodePort: {{.Values.service.nodePort}}
    {{- end }}
```

charts/jaeger-operator/values.yaml
```diff
@@ -23,6 +24,11 @@ rbac:
   pspEnabled: false
   clusterRole: false
 
+service:
+  type: ClusterIP
+  # Specify a specific node port when type is NodePort
+    # nodePort: 32500
```
batazor, line 30: we can align the indents :thinking:

CSP197: I'd aligned it as such, because the service yaml file requires `.Values.service.type.nodePort` to pass in the custom NodePort value.

charts/jaeger-operator/templates/service.yaml
CSP197: Done 👍

CSP197:
@batazor Can you review my updates as per your comments?

charts/jaeger-operator/templates/service.yaml
```diff
@@ -11,7 +11,10 @@ spec:
     port: 8383
     protocol: TCP
     targetPort: 8383
+{{- if and (eq .Values.service.type "NodePort") (.Values.service.nodePort) }}
```
batazor, line 14: The current version will end with an error if you set `nodePort`.

you can try it like this
```
{{ if (and (eq .Values.service.type "NodePort") (not (empty .Values.service.nodePort))) }}
    nodePort: {{.Values.service.nodePort}}
{{- end }}
```

CSP197: This is the same conditional statement as used in the existing `jaeger/templates/query-svc` helm chart: https://github.com/jaegertracing/helm-charts/blob/157d2d5c5d43dfc9bde975ef5b0baf99f42c2c1f/charts/jaeger/templates/query-svc.yaml#L19

batazor:
```diff
-{{ .Values.query.service.nodePort }}
+{{ .Values.service.nodePort }}
```

CSP197: The naming of the variables aside, the logic of the conditional statement is the same and shouldn't result in an error as it is logically equivalent to the one in your suggestion.

batazor: OK, change the path to the variable to the correct path

PR #230, opened 2021-03-30
[Jaeger] Add "automountServiceAccountToken" field to the serviceAccount Templates

mehta-ankit:

#### What this PR does
Allows one to set `automountServiceAccountToken` on the serviceAccounts created by the upstream chart.
https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-the-default-service-account-to-access-the-api-server

N/A

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/values.yaml
+++ b/charts/jaeger/values.yaml
@@ -120,6 +120,8 @@ schema:
     #   memory: 128Mi
   serviceAccount:
     create: true
+    # Explicitly mounts the API credentials for the Service Account
+    automountServiceAccountToken: true
     name:
   podAnnotations: {}
   podLabels: {}
@@ -170,6 +172,8 @@ ingester:
     #   memory: 512Mi
   serviceAccount:
     create: true
+    # Explicitly mounts the API credentials for the Service Account
+    automountServiceAccountToken: true
     name:
   nodeSelector: {}
   tolerations: []
@@ -223,6 +227,8 @@ agent:
     #   memory: 128Mi
   serviceAccount:
     create: true
+    # Explicitly mounts the API credentials for the Service Account
+    automountServiceAccountToken: true
     name:
     annotations: {}
   nodeSelector: {}
@@ -308,6 +314,8 @@ collector:
     #   memory: 512Mi
   serviceAccount:
     create: true
+    # Explicitly mounts the API credentials for the Service Account
+    automountServiceAccountToken: true
     name:
     annotations: {}
   nodeSelector: {}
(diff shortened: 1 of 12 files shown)


charts/jaeger/values.yaml
```diff
@@ -120,6 +120,8 @@ schema:
     #   memory: 128Mi
   serviceAccount:
     create: true
+    # Explicitly mounts the API credentials for the Service Account
+    automountServiceAccountToken: true
```
mehta-ankit, line 124: I believe we can make default to `false`
as mounting the serviceAcount token in a Pod is often not necessary. Pods usually don’t need to communicate with the server Api.
Only reason i made it true is to keep backwards compatibility.

naseemkullah:
Thanks @mehta-ankit, have you run the chart with these set to false? If all works maybe its a better default.

mehta-ankit:
> Thanks @mehta-ankit, have you run the chart with these set to false? If all works maybe its a better default.

@naseemkullah Not yet, will try it and report back.  Trying to add this option to Kafka helm chart as well 😅  (working on too many PR's hehe)

mehta-ankit:
I did install the chart with these change and default value set to `false` for `automountServiceAccountToken` field.
All components pods came up and were running.

As an example for ingester the service account was showing the value set as false. (screenshot below)
image

I also exec'd into the pod and looked at the location where token should have been mounted:
image
There was nothing.

Compared to a Kibana pod i have in my deployment where i need the `automountServiceAccountToken` field to be `true` since it needs a PSP applied to it using a rolebinding (and cluster role).
image

The location marked is the location where secret is mounted : https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod
```
pod is associated with a service account, and a credential (token) for that service account is placed into the filesystem tree of each container in that pod, at /var/run/secrets/kubernetes.io/serviceaccount/token
```

So i think keeping default as `false` for this chart's components should be good IMO.

charts/jaeger/values.yaml
mehta-ankit: resolved in comments on the PR.
See https://github.com/jaegertracing/helm-charts/pull/230#issuecomment-812125874

mehta-ankit:
@naseemkullah @pavelnikolov Can this be reviewed/merged please 🙏🏽 :)

naseemkullah (approved):
Thanks for the details @mehta-ankit !

PR #231, opened 2021-03-31
[jaeger] Adding proxy sidecar to the jaeger-query pod 

OpenetCK:
#### What this PR does
Adding a proxy sidecar (i.e. oauth2-proxy) to the query pod to enable TLS & authentication capabilities.

The idea originated by the following blogpost: https://medium.com/jaegertracing/protecting-jaeger-ui-with-an-oauth-sidecar-proxy-34205cca4bb1

It is up to the user to use any sidecar of their choice, with the necessary volumes mounting (configuration file/certs/key).

Example configuration when using the the [OAuth2-proxy](https://oauth2-proxy.github.io/oauth2-proxy/). 

#####  ConfigMap that contains the OAuth2-Porxy configuration
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: jaeger-oauth-proxy-config
data:
  oauth2-proxy.cfg: |
    provider = "oidc"
    https_address = ":4180"
    upstreams = ["http://localhost:16686"]
    redirect_url = "https://jaeger-svc-domain/oauth2/callback"
    client_id = "jaeger-query"
    client_secret = "xxxxxxxx"
    oidc_issuer_url = "https://keycloak-svc-domain/auth/realms/Default"
    cookie_secret = "RjdmcGVLMVNYaw=="
    cookie_secure = "true"
    provider_ca_files = ["/etc/tls/ca.crt"]
    tls_cert_file = "/etc/tls/tls.crt"
    tls_key_file = "/etc/tls/tls.key"
    email_domains = "*"
    oidc_groups_claim = "groups"
    user_id_claim = "preferred_username"
    skip_provider_button = "true"
```

#####  jaeger Helm chart values
```yaml
query:
  oAuthSidecar:
    enabled: true
    args:
      - --config
    extraConfigmapMounts:
      - name: oauth2-proxy-config
        mountPath: /etc/oauth2-proxy
        configMap: jaeger-oauth-proxy-config
    extraSecretMounts:
      - name: oauth2-proxy-tls-secret
        secretName: oauth2-proxy-tls-secret
        mountPath: /etc/tls
    livenessProbe:
      httpGet:
        path: /ping
        port: oauth-proxy
        scheme: HTTPS
    readinessProbe:
      httpGet:
        path: /ping
        port: oauth-proxy
        scheme: HTTPS
```
Here assumes the `oauth2-proxy-tls-secret` is a valid `kubernetes.io/tls` type of secret contains the `tls.crt`, `tls.key`, and `ca.crt` certificate & keypair.

The jaeger-query service will point to the oauth2-proxy listener port, which has the capability to serve TLS and provide the authentication & authorization features.

N/A

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/charts/jaeger/templates/query-svc.yaml
+++ b/charts/jaeger/templates/query-svc.yaml
@@ -15,7 +15,7 @@ spec:
   - name: query
     port: {{ .Values.query.service.port }}
     protocol: TCP
-    targetPort: query
+    targetPort: {{ ternary "oauth-proxy" "query" .Values.query.oAuthSidecar.enabled }}
 {{- if and (eq .Values.query.service.type "NodePort") (.Values.query.service.nodePort) }}
     nodePort: {{ .Values.query.service.nodePort }}
 {{- end }}
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.22.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
-version: 0.43.3
+version: 0.44.0
 keywords:
   - jaeger
   - opentracing
--- a/charts/jaeger/templates/query-deploy.yaml
+++ b/charts/jaeger/templates/query-deploy.yaml
@@ -119,6 +119,39 @@ spec:
           httpGet:
             path: /
             port: admin
+{{- if .Values.query.oAuthSidecar.enabled }}
+      - name: {{ template "jaeger.agent.name" . }}-oauth2-sidecar
+        image: {{ .Values.query.oAuthSidecar.image }}
+        imagePullPolicy: {{ .Values.query.oAuthSidecar.pullPolicy }}
+        args:
+        {{- range .Values.query.oAuthSidecar.args }}
+          - {{ . }}
+        {{- end }}
+        volumeMounts:
+        {{- range .Values.query.oAuthSidecar.extraConfigmapMounts }}
+          - name: {{ .name }}
+            mountPath: {{ .mountPath }}
+            subPath: {{ .subPath }}
+            readOnly: {{ .readOnly }}
+        {{- end }}
+        {{- range .Values.query.oAuthSidecar.extraSecretMounts }}
+          - name: {{ .name }}
+            mountPath: {{ .mountPath }}
+            subPath: {{ .subPath }}
+            readOnly: {{ .readOnly }}
+        {{- end }}
+        ports:
+          - containerPort: {{ .Values.query.oAuthSidecar.containerPort }}
+            name: oauth-proxy
+        {{- if .Values.query.oAuthSidecar.livenessProbe }}
+        livenessProbe:
+          {{- toYaml .Values.query.oAuthSidecar.livenessProbe | nindent 10 }}
+        {{- end }}
+        {{- if .Values.query.oAuthSidecar.readinessProbe }}
+        readinessProbe:
+          {{- toYaml .Values.query.oAuthSidecar.readinessProbe | nindent 10 }}
+        {{- end }}
+{{- end }}
 {{- if .Values.query.agentSidecar.enabled }}
       - name: {{ template "jaeger.agent.name" . }}-sidecar
         securityContext:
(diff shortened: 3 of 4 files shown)


naseemkullah:
Thanks for submitting this PR!

FWIW I am running oauth2-proxy via https://github.com/oauth2-proxy/manifests/tree/main/helm/oauth2-proxy as a separate service rather than a sidecar for jaeger, but @jpkrohling's article does suggest to use a sidecar approach.

My concern is if other utility sidecars are recommended to be used for other purposes then we will start to get a messy harder to maintain chart.

Firstly I want to confirm that sidecar approach for query is indeed recommended over separate service. @jpkrohling did you consider a separate service approach when writing this article? With a sidecar approach to oauth, you cannot scale out the query UI unless the sidecars use some sort of shared storage for session mgmt.

naseemkullah:
> Can't you use an ingress instead?

Right but ingress should route to an oauth2-proxy (unless the ingress controller has the oauth system built-in), today I am using nginx ing + oauth2-proxy to hit the jaeger UI (and other non protected UIs) more or less as per https://blog.donbowman.ca/2019/02/14/using-single-sign-on-oauth2-across-many-sites-in-kubernetes/

OpenetCK:
The sidecar approach is preferred as we want to have the TLS enabled, to secure the connection everywhere. 
If using service (for the oauth2 proxy), the traffic may travel across different nodes in the cluster, network sniffing seems to be possible. But if it is within the same pod, non-secured plain text only flows within the pod.

It would be like
```
                                                    +------------------+
+--------------+   HTTPS   +------------+  HTTPS    | +--------------+ |
| User browser | --------- |  Ingress   |-----------| | oauth2-proxy | |
+--------------+           +------------+           | +--------------+ |
                                                    |  +-------------+ |
                                                    |  | jaeger-query| |
                                                    |  +-------------+ |
                                                    +------------------+
```

OpenetCK:
Do agree with @naseemkullah that it could get messy when more and more sidecars are added. Was thinking of an approach as simple as just `toYaml` everything under the `spec.template.spec.containers` by introducing a `otherSidecars` property array, but it seems to be too powerful to be hack. 
Then came across the jaeger documentation about the [security](https://www.jaegertracing.io/docs/1.22/security/), on the note of section "brower to ui", since it is documented, then may be it make sense to have a dedicated sidecar just to enable the security feature?
Trying to look for the best practices to achieve this, I guess the best would be the kubernetes webhook method, to modify the deployment and adding whatever required, but that could be ended up requiring more maintenance effort 😐  Any thought?

pavelnikolov:
Although, I like the idea about the sidecar, I'm a bit sceptical to add this here. I'd prefer a single auth-proxy for the cluster. 
https://kubernetes.github.io/ingress-nginx/examples/auth/oauth-external-auth/

OpenetCK:
There is one particular reason sidecar is suggested, which is to prevent the plain-text communicate across different pods (the pod may assigned in different server node).

Ingress & separated deployment for the auth-proxy looks much cleaner, but it couldn't help to solve the Jaeger-UI TLS support requirement. This sidecar approach is same as the security measure taken by the "Client-to-Agent", whereby I quoted from the doc `are meant for trusted environments where the agent is run as a sidecar within the container’s network namespace`

```
                                                                            HTTP                     
+--------------+   HTTPS   +------------+   HTTPS     +--------------+    (insecure)  +--------------+
| User browser | --------- |  Ingress   |------------ | oauth2-proxy | -------------- | jaeger-query |
+--------------+           +------------+             +--------------+    *sniff***   +--------------+
```

naseemkullah (approved):
Thanks @OpenetCK  for the explanation, FWIW service meshes come in handy to enable mTLS across meshed services.

lgtm, @pavelnikolov any objections?

charts/jaeger/templates/query-svc.yaml
```diff
@@ -15,7 +15,7 @@ spec:
   - name: query
     port: {{ .Values.query.service.port }}
     protocol: TCP
-    targetPort: query
+    targetPort: {{ ternary "oauth-proxy" "query" .Values.query.oAuthSidecar.enabled }}
```
naseemkullah, line 18: oh nice

pavelnikolov:
No objections. Feel free to add it. I just solve this problem differently.

PR #245, opened 2021-05-08
[jaeger]: Add option to use custom auth proxy with jaeger query 

nflaig:
#### What this PR does

The PR adds the option to use a custom auth proxy with jaeger query by adding values to add additional `sidecars` to the jeager query pods and a value to customize the `targetPort` of the service. I also added another value to add `extraVolumes` which allows for more flexibility when configuring `sidecars`. Also even though the main use case for `sidecars` is probably to add a auth proxy it is not explicitly limited to that.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/values.yaml
+++ b/charts/jaeger/values.yaml
@@ -401,6 +401,8 @@ query:
     # List of IP ranges that are allowed to access the load balancer (if supported)
     loadBalancerSourceRanges: []
     port: 80
+    # Specify a custom target port (e.g. port of auth proxy)
+    # targetPort: 8080
     # Specify a specific node port when type is NodePort
     # nodePort: 32500
   ingress:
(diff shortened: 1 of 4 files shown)


charts/jaeger/values.yaml
```diff
@@ -445,6 +447,16 @@ query:
   #   subPath: ""
   #   configMap: jaeger-config
   #   readOnly: true
+  extraVolumes: []
+  # Array to add extra volumes
```
naseemkullah, line 451:
```suggestion
```

The comment provides no value imho

charts/jaeger/values.yaml
```diff
@@ -445,6 +447,16 @@ query:
...
+  sidecars: {}
+  ## Add sidecar containers to the jaeger query pods
+  ## sidecars:
+  ##   - name: your-image-name
+  ##     image: your-image
+  ##     ports:
+  ##       - name: portname
+  ##         containerPort: 1234
```
naseemkullah, line 459:
```suggestion
  sidecars: []
  ##   - name: your-image-name
  ##     image: your-image
  ##     ports:
  ##       - name: portname
  ##         containerPort: 1234
```

naseemkullah (changes requested):

nflaig:
@naseemkullah Thanks for the review! I applied the suggested changes and rebased against main

naseemkullah (approved):
Thanks @nflaig

PR #256, opened 2021-06-14
[jaeger] add environment variables to jaeger oauth sidecar

jdgeisler:
#### What this PR does
This PR allows the passing of environment variables to the oauth2-proxy sidecar for Jaeger.

The environment variables are passed in values.yaml under the `extraEnv` key, and are then passed into the oauth2-sidecar container.

The reasoning for this is so that we can use kubernetes secrets as environment variables within the oauth proxy ConfigMap.

Example:
```
extraEnv:
  - name: test-env
    valueFrom:
      secretKeyRef:
        name: test-secret
        key: test-key
```
```
oAuthSidecar:
  enabled: true
  args:
    - --config
    - --client-secret
    - "$(test-env)"
```

N/A

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.22.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
-version: 0.46.0
+version: 0.46.1
 keywords:
   - jaeger
   - opentracing
--- a/charts/jaeger/templates/query-deploy.yaml
+++ b/charts/jaeger/templates/query-deploy.yaml
@@ -132,6 +132,10 @@ spec:
         {{- range .Values.query.oAuthSidecar.args }}
           - {{ . }}
         {{- end }}
+        {{- if .Values.query.oAuthSidecar.extraEnv }}
+        env:
+          {{- toYaml .Values.query.oAuthSidecar.extraEnv | nindent 10 }}
+        {{- end }}
         volumeMounts:
         {{- range .Values.query.oAuthSidecar.extraConfigmapMounts }}
           - name: {{ .name }}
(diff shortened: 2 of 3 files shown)


dicolasi:
This is a very important PR, since the previous one added secrets in clear with the CM. Would it be possible to get this merged?

mehta-ankit:
@jdgeisler thanks for the contribution.
For my curiosity, can you point me to the docs which explain this `client_secret = "#{ENV['test-env']}"`

jdgeisler:
> @jdgeisler thanks for the contribution.
> For my curiosity, can you point me to the docs which explain this `client_secret = "#{ENV['test-env']}"`

So actually looks like I referenced a wrong example for configuring the environment variables in oauth2. 
`client_secret = "#{ENV['test-env']}"` would not work in this case, as it is actually a ruby expression that would work with fluentd.

The functionality in this PR still works though, as instead the environment variable could be referenced in the `args` section of the `oAuthSidecar` like this example below.

```
oAuthSidecar:
  enabled: true
  args:
    - --config
    - --client-secret
    - "$(test-env)"
```

dicolasi:
Yes, that is the right way @jdgeisler

dicolasi:
I can reply to the question **Which issue this PR fixes**. 
This PR would avoid having this in the configMap:

    client_secret = "xxxxxxxx"

as per PR -> https://github.com/jaegertracing/helm-charts/pull/231

mehta-ankit:
Thank you for all that information.
Found creation and usage of env var here as well: https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config

mehta-ankit (approved):

PR #259, opened 2021-06-18
[jaeger] Add a configmap for the jaeger oauth sidecar

jdgeisler:
#### What this PR does

This PR adds an optional configmap to be deployed with the oauth2-proxy-sidecar. 

In values.yaml, if the `config` key in the `oAuthSidecar` yaml is added, then the `oauth-sidecar-configmap.yaml` is deployed with the specified data. If the `config` key is not passed, then the `oauth-sidecar-configmap` will not be deployed.

If the config is passed, it is then automatically added as a volume and mounted as `jaeger-oauth-configuration`under the path `/etc/oauth2-proxy`

example values.yaml:
```
oAuthSidecar:
  enabled: true
  pullPolicy: IfNotPresent
  containerPort: 4180
  args:
    - --config
  extraEnv: []
  extraConfigmapMounts: []
  extraSecretMounts: []
  config: |-
    provider = "oidc"
    https_address = ":4180"
    upstreams = ["http://localhost:16686"]
    redirect_url = "https://jaeger-svc-domain/oauth2/callback"
    client_id = "jaeger-query"
    oidc_issuer_url = "https://keycloak-svc-domain/auth/realms/Default"
    cookie_secure = "true"
    email_domains = "*"
    oidc_groups_claim = "groups"
    user_id_claim = "preferred_username"
    skip_provider_button = "true"
```

Per initial PR https://github.com/jaegertracing/helm-charts/pull/231 for the oauth sidecar, the configmap had to be deployed manually outside of the jaeger helm chart.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/README.md
+++ b/charts/jaeger/README.md
@@ -264,6 +264,48 @@ helm install jaeger jaegertracing/jaeger \
   --set storage.kafka.topic=<TOPIC>
 ```
 
+## oAuth2 Sidecar
+If extra protection of the Jaeger UI is needed, then the oAuth2 sidecar can be enabled in the Jaeger Query. The oAuth2
+sidecar acts as a security proxy in front of the Jaeger Query service and enforces user authentication before reaching
+the Jaeger UI. This method can work with any valid provider including Keycloak, Azure, Google, GitHub, and more.
+
+Content of the `jaeger-values.yaml` file:
+
+```YAML
+query:
+  enabled: true
+  oAuthSidecar:
+    enabled: true
+    image: quay.io/oauth2-proxy/oauth2-proxy:v7.1.0
+    pullPolicy: IfNotPresent
+    containerPort: 4180
+    args:
+      - --config
+      - /etc/oauth2-proxy/oauth2-proxy.cfg
+      - --client-secret
+      - "$(client-secret)"
+    extraEnv:
+      - name: client-secret
+        valueFrom:
+          secretKeyRef:
+            name: client-secret
+            key: client-secret-key
+    extraConfigmapMounts: []
+    extraSecretMounts: []
+    config: |-
+      provider = "oidc"
+      https_address = ":4180"
+      upstreams = ["http://localhost:16686"]
+      redirect_url = "https://jaeger-svc-domain/oauth2/callback"
+      client_id = "jaeger-query"
+      oidc_issuer_url = "https://keycloak-svc-domain/auth/realms/Default"
+      cookie_secure = "true"
+      email_domains = "*"
+      oidc_groups_claim = "groups"
+      user_id_claim = "preferred_username"
+      skip_provider_button = "true"
+```
+
 ## Installing extra kubernetes objects
 
 If additional kubernetes objects need to be installed alongside this chart, set the `extraObjects` array to contain 
--- a/charts/jaeger/templates/oauth-sidecar-configmap.yaml
+++ b/charts/jaeger/templates/oauth-sidecar-configmap.yaml
@@ -1,0 +1,12 @@
+{{- if and .Values.query.oAuthSidecar.enabled .Values.query.oAuthSidecar.config}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "jaeger.fullname" . }}-oauth-configuration
+  labels:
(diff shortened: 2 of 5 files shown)


charts/jaeger/templates/oauth-sidecar-configmap.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.query.oAuthSidecar.enabled }}
```
mehta-ankit, line 1: can we maybe do this:
```
{{- if and .Values.query.oAuthSidecar.enabled .Values.query.oAuthSidecar.config}}
```
🤔

jdgeisler: Good idea, will change this now 👍

mehta-ankit:
@jdgeisler If you don't mind, can you please combine these 5 commits into 1 or 2 meaningful commits (which add the functionality).
Also I was looking at the docs here: https://github.com/jaegertracing/helm-charts/tree/main/charts/jaeger and there is no mention of oauth sidecar for query. Can we please add some information here for the users.
Thanks

jdgeisler:
> https://github.com/jaegertracing/helm-charts/tree/main/charts/jaeger

That is no problem at all. I will combine those 5 commits into one and then add another commit for updating the documentation.

jdgeisler:
> @jdgeisler If you don't mind, can you please combine these 5 commits into 1 or 2 meaningful commits (which add the functionality).
> Also I was looking at the docs here: https://github.com/jaegertracing/helm-charts/tree/main/charts/jaeger and there is no mention of oauth sidecar for query. Can we please add some information here for the users.
> Thanks

That is no problem at all. I will combine those 5 commits into one and then add another commit for updating the documentation.

mehta-ankit:
One more question @jdgeisler instead of asking users to add the extra configmap mount
```
extraConfigmapMounts:
    - name: jaeger-oauth-configuration
      mountPath: /etc/oauth2-proxy
      configMap: jaeger-oauth-configuration
```
Can't we add it when config is added in the values file. Same as we do for jaeger query 🤔 . why add another step for the user.
Thoughts ?

jdgeisler:
> One more question @jdgeisler instead of asking users to add the extra configmap mount
> 
> ```
> extraConfigmapMounts:
>     - name: jaeger-oauth-configuration
>       mountPath: /etc/oauth2-proxy
>       configMap: jaeger-oauth-configuration
> ```
> 
> Can't we add it when config is added in the values file. Same as we do for jaeger query 🤔 . why add another step for the user.
> Thoughts ?

Oh I did not think about that, but I like that idea a lot! Makes a lot more sense for the user. I can push those changes out shortly.

jdgeisler:
@mehta-ankit Thank you for the suggestions! I believe everything should be changed now. Please let me know if anything needs to be added or fixed in the docs.

charts/jaeger/README.md
```diff
@@ -264,6 +264,48 @@ helm install jaeger jaegertracing/jaeger \
...
+query:
+  enabled: true
+  oAuthSidecar:
+    enabled: true
+    image: quay.io/oauth2-proxy/oauth2-proxy:v7.1.0
+    pullPolicy: IfNotPresent
+    containerPort: 4180
+    args:
+      - --config
+      - /etc/oauth2-proxy/oauth2-proxy.cfg
```
mehta-ankit, line 284: this should point to the path of the volumeMount right ? 🤔 
```suggestion
      - /etc/oauth2-proxy
```

jdgeisler: I believe it should actually point to the data key within the ConfigMap, which is `oauth2-proxy.cfg`. My team currently has the path set as `/etc/oauth2-proxy/oauth2-proxy.cfg` and it is working.

Per the oauth2 docs, the `--config` option takes the path to the config file https://oauth2-proxy.github.io/oauth2-proxy/docs/configuration/overview#command-line-options

mehta-ankit: Yes my bad 🤦🏽‍♂️ . It should be full path to the file. Not sure what i was thinking.

mehta-ankit (approved):
LGTM

@naseemkullah @pavelnikolov Can 1 of you take a look at this change as well. I feel good about it but wanted another set of eyes.

PR #260, opened 2021-06-22
[jaeger-operator]: fix the problem with upgrade chart

batazor:

Resolves the conflict of immutable labels when upgrading from an older version

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger-operator/Chart.yaml
+++ b/charts/jaeger-operator/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: jaeger-operator Helm chart for Kubernetes
 name: jaeger-operator
-version: 2.21.4
+version: 2.21.5
 appVersion: 1.22.1
 home: https://www.jaegertracing.io/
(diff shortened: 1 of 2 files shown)


stevehipwell:
Wouldn't it be better to handle this as a value to conditionally exclude the new (but correct) label for existing chart upgrades? Or maybe just some docs on deleting the operator deployment and re-upgrading?

batazor:
> Wouldn't it be better to handle this as a value to conditionally exclude the new (but correct) label for existing chart upgrades? Or maybe just some docs on deleting the operator deployment and re-upgrading?

@stevehipwell I agree, corrected it so that it would not break the old settings, but work on new ones. What do you think?

stevehipwell:
@batazor I think you'd need to add an actual value to _values.yaml_ such as `legacyLabels` instead of using `.Release.IsUpgrade`, as the current code would break for new installs on the first upgrade. I also think that this should probably be a new major version (or at least minor one) so the messaging would be **If you installed before v3 you need to set `legacyLabels: true` or delete the deployment when upgrading to a v3 version.**.

PR #263, opened 2021-06-30
[jaeger-operator] Enforce default container security policy

majidazimi:
Since upstream docker image is already using non root user, I thought it might be good to enforce security context at container level in the chart to prevent accidental problems. 

Does it make sense to make this configurable? jaeger-operator will always run under non root and it will not modify local file system anyways. So I hardcoded all the parameters.


majidazimi:
lint-test stage has failed without a clear error message:
```
 jaeger-operator => (version: "2.22.1", path: "charts/jaeger-operator") > Error waiting for process: exit status 1
```

mehta-ankit:
> lint-test stage has failed without a clear error message:
> 
> ```
>  jaeger-operator => (version: "2.22.1", path: "charts/jaeger-operator") > Error waiting for process: exit status 1
> ```

I saw similar failures on the job for jaeger helm chart PR's. Pod is failing to be scheduled because node has a taint.
```
64s         Warning   FailedScheduling    pod/jaeger-operator-7n3zv2gdkq-7c7c5bcc8-22fpg                default-scheduler       0/1 nodes are available: 1 node(s) had taint {node.kubernetes.io/not-ready: }, that the pod didn't tolerate.   10m          8       jaeger-operator-7n3zv2gdkq-7c7c5bcc8-22fpg.168d5b94a8712b0c
```

majidazimi:
I fixed the version conflict by bumping chart version. Would you please take a look into this?

charts/jaeger-operator/templates/deployment.yaml
```diff
@@ -36,6 +36,11 @@ spec:
         - name: {{ include "jaeger-operator.fullname" . }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          securityContext:
```
chgl, line 39: If you don't mind a small comment: I personally prefer the flexibility of setting the entire securityContext via values.yaml, e.g.:

```yaml
# values.yaml:
securityContext:
  allowPrivilegeEscalation: false
  capabilities:
    drop:
      - ALL
  readOnlyRootFilesystem: true
  runAsNonRoot: true
  privileged: false
  runAsUser: 65532
```
with deployment.yaml being:

```yaml
          securityContext:
            {{- toYaml .securityContext | nindent 12 }}
```

majidazimi: Sure. However, since there is already `securityContext` at pod level, it would be better to add `containerSecurityContext` in `values.yml`

PR #265, opened 2021-07-05
[jaeger] Make service port of collector ingress configurable

### LINKED ISSUE #264: Collector ingress: make service port configurable
## Requirement - what kind of business use case are you trying to solve?

I would like to be able to configure which collector service port is bound to the collector ingress.

## Problem - what in Jaeger blocks you from solving the requirement?

Currently the ingress of the collector, when enabled, always uses the "http" port of its service:

https://github.com/jaegertracing/helm-charts/blob/main/charts/jaeger/templates/collector-ing.yaml#L2

## Proposal - what do you suggest to solve the problem or improve the existing situation?

The service port that the collector ingress uses should be configurable. For example, we would like to use Zipkin. There could even per a port configuration per host so that different protocols can be exposed with different host names.

## Any open questions to address

-

hoepfnerj:
Signed-off-by: Jonas Höpfner <jonashoepfner@gmx.de>

#### What this PR does

This PR changes the collector ingress and makes the bound service port configurable.

- fixes #264

#### Checklist

- [X] Chart Version bumped
- [X ] Title of the PR starts with chart name (`[jaeger]` or `[jaeger-operator]`)


locmai:
It would be great if this could get merged. It's quite useful as we are moving toward the gRPC ingress now instead of the HTTP.

Could someone take a look?

charts/jaeger/templates/collector-ing.yaml
```diff
@@ -14,14 +13,14 @@ metadata:
...
-    - host: {{ $host }}
+    {{- range .Values.collector.ingress.hosts }}
+    - host: {{ .host }}
       http:
         paths:
           - path: {{ $basePath }}
             backend:
               serviceName: {{ template "jaeger.collector.name" $ }}
-              servicePort: {{ $servicePort }}
+              servicePort: {{ .servicePort }}
```
naseemkullah, line 23: Would it make sense to fallback to `.Values.collector.service.http.port` which is the current value for folks who are not explicitly setting the servicePort?

hoepfnerj: That would be an option. However, personally I think it should be a conscious decision which port to use. Since the value for `host` must be configured anyway and the `servicePort` value is right next to it (with an example given) I think there is not that much benefit for a fallback.

naseemkullah: I would agree if the ingress collector was not already implemented, but my concern is to not introduce breakage to users who are using the collector ingress today.

hoepfnerj: Ok, this makes sense. I will take a look at it tonight.

hoepfnerj: I changed it so that `.Values.collector.service.http.port` is now the fallback.

PR #258, opened 2021-06-17
[jaeger] Reordering HPA metrics section

saikalyan551:
Signed-off-by: sai kalyan <svytla@axon.com>

#### What this PR does
This PR fixes the HPA metrics list in a way that the order will be the same as how the HPA controller is saving it. There is a problem while using GitOps style of CD's like Argocd,harness...etc. The HPA controller is swaping the order of metrics section which is causing difference while using GitOps models CD's.
 Example of HPA for Argocd:
 HPA.yaml
```
spec:
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name:  jaeger-collector
  minReplicas: 2
  maxReplicas: 10
  metrics:
    - type: Resource
      resource:
        name: cpu
        targetAverageUtilization: 90
    - type: Resource
      resource:
        name: memory
        targetAverageUtilization: 90

```
 Live Manifest that its applying:
```
  name: jaeger-collector
  namespace: jaeger-components
spec:
  maxReplicas: 10
  metrics:
    - resource:
        name: memory
        targetAverageUtilization: 90
      type: Resource
    - resource:
        name: cpu
        targetAverageUtilization: 90
      type: Resource
  minReplicas: 2
```
 Clearly, you see a difference in reordering of metrics section, which is causing outofsync kind of issues. So, if we can just reorder the memory & CPU resource it can solve it.

#### Which issue this PR fixes

https://github.com/argoproj/argo-cd/issues/1079

*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format,
will close that issue when PR gets merged)*

- fixes #

#### Checklist

- [x] [DCO](https://github.com/jaegertracing/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Commits are [GPG signed](https://docs.github.com/en/github/authenticating-to-github/about-commit-signature-verification)
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (`[jaeger]` or `[jaeger-operator]`)

PR #267, opened 2021-07-07
[jaeger]: Add activeDeadlineSeconds value to all jobs

nflaig:
#### What this PR does

Adds `activeDeadlineSeconds` value to all jobs to allow chart users to set a job deadlines which is currently required to terminate a job if sidecars are used.  This issue is also described [here](https://github.com/kubernetes/kubernetes/issues/25908) and there is not yet a proper solution for this.

An example use case would be if you are running elasticsearch in a service mesh which means in order for the job containers to send requests to elasticsearch it is required to also add them to the mesh hence injecting additional sidecars. The jobs will not finish by themselves since the sidecars are not shutting down which is a big issue for the hook/init job but also for all cronjobs because the pods are not cleaned up after the jobs finishes.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.22.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
-version: 0.46.3
+version: 0.46.4
 keywords:
   - jaeger
   - opentracing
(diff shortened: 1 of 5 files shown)


mehta-ankit (approved):
LGTM
Can you rebase with main branch please.
FYI: https://github.com/jaegertracing/helm-charts/pull/258#issuecomment-875707234

nflaig:
@mehta-ankit Thanks for the review! I rebased the branch against main

mehta-ankit:
@nflaig You need to GPG sign the commit. Merge is blocked till then.

nflaig:
@mehta-ankit sorry I missed that I think the GPG signature was lost when rebasing or amending the commit

PR #269, opened 2021-07-13
[jaeger]: Only add ServiceMonitor resources if crd exists

nflaig:
#### What this PR does

The PR adds a check to all `ServiceMonitor` to only add them if the crd (`monitoring.coreos.com/v1` api) exists. This allows to set `<component>.serviceMonitor.enabled` to `true` while ensuring that the installation of the whole chart does not fail just because the `ServiceMonitor` resource can not be created.

Potential consideration is that now if `<component>.serviceMonitor.enabled`  is set to `true` but the `monitoring.coreos.com/v1` does not exist it will silently not create the resources instead of throwing an error when installing the chart. Not sure if this should be considered a disadvantage.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.22.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
-version: 0.46.4
+version: 0.46.5
 keywords:
   - jaeger
   - opentracing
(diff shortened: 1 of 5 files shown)


mehta-ankit:
IMO I haven't seen this pattern on helm charts before and i feel the chart should be generic enough and care about resources it needs to create vs capabilities of the user's K8s cluster. Some people have alpha feature gates enabled , some only have beta one's enabled. 
Also your point about the `failure` being silent now is concerning.

@naseemkullah thoughts ?

nflaig:
@mehta-ankit I saw this pattern in the [fluentd helm chart](https://github.com/fluent/helm-charts/blob/main/charts/fluentd/templates/servicemonitor.yaml#L1) and I find it quite useful.

naseemkullah:
I think I would prefer to have it fail and force me to be explicit about enabling/disabling the service monitors. But Nico you mentioned that you find this useful could you elaborate as to why

nflaig:
@naseemkullah we are using a helm chart which bundles all monitoring tools such as jaeger, prometheus operator, grafana, fluentd, etc. We deploy this chart to multiple systems which might not yet have prometheus operator enabled but it quite nice to still have all servie monitors enabled in the helm chart defautlt values so we don't need to explicitly enable them on the system specific values configuration. Without a check if the api is available this can not be done because the deployed would fail so we need to maintain more configuration values for systems with and without prometheus operator.

I guess this is a bit specific use case and probably makes sense to close this one. Thanks for the review and feedback

PR #275, opened 2021-08-06
[jaeger] use networking.k8s.io/v1 for ingreses if available

### LINKED ISSUE #273: Deploying ingress will fail with kubernetes 1.22
**Describe the bug**
The ingress templates for jaeger use `networking.k8s.io/v1beta1` as apiVersion whose support will be dropped with kubernetes 1.22 (aka the next version of kubernetes) [1].

There are basically two options:

**1. Drop support for v1beta1**

Since this will lock out users using kubernetes <= 1.18 (and I guess there are still a few aorund) this is not really an option.

**2. Define ingress based on capabilities**

Helm allows a check if the used kubernetes version supports a specific capability via `Capabilities.APIVersions.Has` [2]. As there is also a slightly different syntax between v1beta1 and v1 a few more if/elses need to be added.

It would look then something like this:

```
{{- $useV1 := .Capabilities.APIVersions.Has "networking.k8s.io/v1" -}}
{{- if $useV1 }}
apiVersion: networking.k8s.io/v1
{{ else }}
apiVersion: networking.k8s.io/v1beta1
{{ end }}
```

It will make the ingress templates more complex but I think this is the way to go.

If there are no objections or different ideas I will start work on this in the next days.

[1] https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122
[2] https://helm.sh/docs/chart_template_guide/builtin_objects/

bender316:
Signed-off-by: bender316 <bender316@gmail.com>

#### What this PR does

Starting with kubernetes v1.22 the old api version "networking.k8s.io/v1beta1" for ingresses is no longer supported. This PR supports now both flavors.

- fixes #273 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.22.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
-version: 0.46.4
+version: 0.46.5
 keywords:
   - jaeger
   - opentracing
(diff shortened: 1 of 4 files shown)


naseemkullah (approved):
Thanks @bender316

PR #278, opened 2021-08-24
[jaeger] Include bitnami's common chart to fix Ingress v1/v1beta1 distinction

### LINKED ISSUE #277: Deploying with ingress fails
**Describe the bug**
Since Helm Chart commit ac557aa198d50910b6c4f71f76a54d60de26d63c the deployment fails for us, as we don't have `networking.k8s.io/v1/Ingress` but we do have `networking.k8s.io/v1/NetworkPolicy`.

**To Reproduce**
1. Have a cluster without `networking.k8s.io/v1/Ingress` but with something else in apiVersion `networking.k8s.io/v1`, e.g. `NetworkPolicy`
2. Try to deploy with ingress enabled
3. See error message

**Expected behavior**
Either a better check, which I don't think is possible with Helm?, or an additional flag whether to use `networking.k8s.io/v1` for the ingress

**Screenshots**

**Version (please complete the following information):**
 - OS: N.A.
 - Jaeger version: N.A.
 - Deployment: Kubernetes
 - Kubernetes Version: v1.18.8
 - Helm Chart version: 0.46.5

**What troubleshooting steps did you try?**
I looked through the source to find a toggle or something like that

**Additional context**
Log;

```log
installation failed: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "Ingress" in version "networking.k8s.io/v1"
```

cwrau:
Replace Ingress `v1`/`v1beta1` logic

Signed-off-by: Chris Werner Rau <cwrau@cwrcoding.com>

#### What this PR does

Uses bitnami's `common` chart to replace the Ingress `v1`/`v1beta1` logic, as that had a flaw

- fixes #277

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.22.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
-version: 0.46.5
+version: 0.46.6
 keywords:
   - jaeger
   - opentracing
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -37,3 +37,6 @@ dependencies:
     version: ^12.16.0
     repository: https://charts.bitnami.com/bitnami
     condition: provisionDataStore.kafka
+  - name: common
+    repository: https://charts.bitnami.com/bitnami
+    version: 1.8.0
--- a/charts/jaeger/templates/collector-ing.yaml
+++ b/charts/jaeger/templates/collector-ing.yaml
@@ -1,12 +1,7 @@
-{{- $useNetworkingV1 := .Capabilities.APIVersions.Has "networking.k8s.io/v1" -}}
 {{- if .Values.collector.ingress.enabled -}}
-{{- $servicePort := .Values.collector.service.http.port -}}
-{{- $basePath := .Values.collector.basePath -}}
-{{- if $useNetworkingV1 }}
-apiVersion: networking.k8s.io/v1
-{{- else }}
-apiVersion: networking.k8s.io/v1beta1
-{{- end }}
+  {{- $servicePort := .Values.collector.service.http.port -}}
+  {{- $basePath := .Values.collector.basePath -}}
+apiVersion: {{ include "common.capabilities.ingress.apiVersion" $ }}
 kind: Ingress
 metadata:
   name: {{ template "jaeger.collector.name" . }}
@@ -15,7 +10,7 @@ metadata:
     app.kubernetes.io/component: collector
   {{- if .Values.collector.ingress.annotations }}
   annotations:
-    {{- toYaml .Values.collector.ingress.annotations | nindent 4 }}
+  {{- toYaml .Values.collector.ingress.annotations | nindent 4 }}
   {{- end }}
 spec:
   rules:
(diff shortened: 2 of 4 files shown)


charts/jaeger/templates/collector-ing.yaml
```diff
@@ -24,21 +19,13 @@ spec:
       http:
         paths:
           - path: {{ $basePath }}
-            {{- if $useNetworkingV1 }}
+            {{- if (include "common.ingress.supportsPathType" $) }}
             pathType: ImplementationSpecific
```
cwrau, line 23: I'm unsure about this

Before you only set the `pathType` when you had `v1` and in the same logic used the new ingress.

Now this will be set even on 1.18, as that k8s version already supports the `pathType`

charts/jaeger/templates/collector-ing.yaml
```diff
@@ -15,7 +10,7 @@ metadata:
     app.kubernetes.io/component: collector
   {{- if .Values.collector.ingress.annotations }}
   annotations:
-    {{- toYaml .Values.collector.ingress.annotations | nindent 4 }}
+  {{- toYaml .Values.collector.ingress.annotations | nindent 4 }}
```
naseemkullah, line 13: Could you keep this indented? it reads better imho

charts/jaeger/templates/collector-ing.yaml
```diff
@@ -1,12 +1,7 @@
...
 {{- if .Values.collector.ingress.enabled -}}
-{{- $servicePort := .Values.collector.service.http.port -}}
-{{- $basePath := .Values.collector.basePath -}}
-{{- if $useNetworkingV1 }}
-apiVersion: networking.k8s.io/v1
-{{- else }}
-apiVersion: networking.k8s.io/v1beta1
-{{- end }}
+  {{- $servicePort := .Values.collector.service.http.port -}}
+  {{- $basePath := .Values.collector.basePath -}}
```
naseemkullah, line 3:
```suggestion
{{- $servicePort := .Values.collector.service.http.port -}}
{{- $basePath := .Values.collector.basePath -}}
```

charts/jaeger/templates/hotrod-ing.yaml
```diff
@@ -1,46 +1,33 @@
...
   name: {{ include "jaeger.fullname" . }}-hotrod
   labels:
     {{- include "jaeger.labels" . | nindent 4 }}
     app.kubernetes.io/component: hotrod
-{{- if .Values.hotrod.ingress.annotations }}
+  {{- if .Values.hotrod.ingress.annotations }}
   annotations:
-    {{- toYaml .Values.hotrod.ingress.annotations | nindent 4 }}
-{{- end }}
+  {{- toYaml .Values.hotrod.ingress.annotations | nindent 4 }}
```
naseemkullah, line 14: same

charts/jaeger/templates/query-ing.yaml
```diff
@@ -15,7 +10,7 @@ metadata:
     app.kubernetes.io/component: query
   {{- if .Values.query.ingress.annotations }}
   annotations:
-    {{- toYaml .Values.query.ingress.annotations | nindent 4 }}
+  {{- toYaml .Values.query.ingress.annotations | nindent 4 }}
```
naseemkullah, line 13: same

charts/jaeger/templates/collector-ing.yaml
naseemkullah: @bender316 could we get your thoughts on this (since you implemented the original logic)?

charts/jaeger/templates/collector-ing.yaml
cwrau: Of course, this is IntelliJ's style, but I can revert to your style 👍️

naseemkullah (approved):
Going to approve now but will give @bender316 a few days to respond about the `pathType` being set as of k8s 1.18

Thanks @cwrau, great to offload this logic to an external chart!

pavelnikolov (approved):

charts/jaeger/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: 1.22.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
-version: 0.46.5
```
pavelnikolov, line 6: (nitpick) I'd bump it to 0.47.0.

cwrau: Really? Currently, I can't deploy this chart to our cluster because of #277.

That's why I thought this change to be a bugfix rather than a feature, but I can change it if you want

charts/jaeger/templates/collector-ing.yaml
naseemkullah: Seems like 1.18 already has issues with the current logic according to https://github.com/jaegertracing/helm-charts/pull/279

PR #279, opened 2021-08-31
[jaeger] Check for using "networking.k8s.io/v1" is incorrect

hardik-fareye:
the check for 
you have applied:  {{- $useNetworkingV1 := .Capabilities.APIVersions.Has "networking.k8s.io/v1" -}}
this is true for older versions of k8s as well but in older version, "networking.k8s.io/v1" does not have Ingress.
please apply the version check:

kubernetes version 1.19+ has ingress in "networking.k8s.io/v1"

- fixes #

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/jaeger/templates/query-ing.yaml
+++ b/charts/jaeger/templates/query-ing.yaml
@@ -2,7 +2,7 @@
 {{- if .Values.query.ingress.enabled -}}
 {{- $servicePort := .Values.query.service.port -}}
 {{- $basePath := .Values.query.basePath -}}
-{{- if $useNetworkingV1 }}
+{{- if semverCompare ">=1.19" .Capabilities.KubeVersion.Version -}}
 apiVersion: networking.k8s.io/v1
 {{- else }}
 apiVersion: networking.k8s.io/v1beta1


naseemkullah:
Thanks for raising this issue @hardik-fareye, we've just released 0.46.6  where @cwrau imported bitnami's https://github.com/bitnami/charts/tree/master/bitnami/common to standardize on api versioning logic, could you please try it out, if it still fails please rebase to get the latest changes. Cheers

mehta-ankit:
@hardik-fareye Can you confirm if this is still an issue for you with latest version of the chart ?? If not i would like to close this issue.

charts/jaeger/templates/query-ing.yaml
```diff
@@ -2,7 +2,7 @@
 {{- if .Values.query.ingress.enabled -}}
 {{- $servicePort := .Values.query.service.port -}}
 {{- $basePath := .Values.query.basePath -}}
-{{- if $useNetworkingV1 }}
+{{- if semverCompare ">=1.19" .Capabilities.KubeVersion.Version -}}
```
pavelnikolov, line 5: What is wrong with this check?
```
{{- $useNetworkingV1 := .Capabilities.APIVersions.Has "networking.k8s.io/v1" -}}
```

pavelnikolov: It might be related to a bug in `helm` https://github.com/databus23/helm-diff/issues/165
Which version of helm are you using? Can you try upgrading and see if you'd have the same issue?
I'd rather not use version comparisons. We should always check if the appropriate API version is available regardless of the Kubernetes version being used.

pavelnikolov:
I'm closing this PR due to inactivity. Please, feel free to reopen.

PR #298, opened 2021-10-20
[Jaeger] Add support for individual image tags

sharninder:
#### What this PR does

The current charts only have support for adding the tag of a docker image at the very top. That means that all images being used in the chart for the various jaeger components must have the same tag. This is sometimes not a good idea. This PR attempts to add support for having individual tags for each component where it makes sense.


sharninder:
@pavelnikolov should be signed now, no ? Sorry I'm new to git signing

pavelnikolov:
@sharninder only the last commit is signed. Can you, please, squash the commits into one?

PR #309, opened 2021-11-01
[jaeger] chore: adding collector supports install in new namespace

vinisdl:
Signed-off-by: Marco Dalalba <vinisdl@hotmail.com>
#### What this PR does
Allows you to install jaeger in a namespace other than the default.

#### Checklist

- [x]  [DCO] (https://github.com/jaegertracing/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed

- [ ] Commitments are [signed by GPG] (https://docs.github.com/en/github/authenticating-to-github/about-commit-signature-verification)
- [ ] The graphic version has collapsed
- [x] The PR title starts with the chart name (`[jaeger]` or `[jaeger-operator]`)

### DIFF
--- a/charts/jaeger/templates/_helpers.tpl
+++ b/charts/jaeger/templates/_helpers.tpl
@@ -230,6 +230,14 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- end -}}
 
 {{/*
+Create a fully qualified collector host.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+*/}}
+{{- define "jaeger.collector.host" -}}
+{{- printf "%s.%s" (include "jaeger.fullname" .) .Release.Namespace  | trunc 63 -}}
+{{- end -}}
+
+{{/*
 Create a fully qualified ingester name.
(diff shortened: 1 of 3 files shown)


mehta-ankit:
@vinisdl Can you please clarify in some detail what is the issue that this PR is resolving. I am not sure i fully understand that.
I feel like we are going from 
`<jaeger.fullname>-collector` to `<jaeger.fullname>.namesapce` ?

vinisdl:
If you try to install jaeger in a namespace other than the default you will get an error as there is no service

jaeger-collector.default.svc.cluster.local (ex)

you need to add the namespace to work correctly

With this pull request you fix this problem

naseemkullah:
What I don't quite understand is the query and agent are deployed in the same namespace as the collector so using just the name should resolve 🤔

mehta-ankit:
> What I don't quite understand is the query and agent are deployed in the same namespace as the collector so using just the name should resolve 🤔

I wanted to mention the same, after checking my own deployment. This is how the env variable is set for my deployment
```
- env:
        - name: REPORTER_GRPC_HOST_PORT
          value: jaeger-collector:14250
```
And it works.

PR #323, opened 2021-12-09
[jaeger] use jaeger.image.tag for es-index-cleaner tag

manuelbonk:
Signed-off-by: Manuel Bonk <gh@mbonk.de>

#### What this PR does

This PR sets the default image tag for `jaeger-es-index-cleaner` cronjob  from `latest` to  `jaeger.image.tag` (see https://github.com/jaegertracing/helm-charts/blob/main/charts/jaeger/templates/es-index-cleaner-cronjob.yaml#L48) In case you want to override the tag you can still do so manually, but I think it shouldn't be set by default to `latest`.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.28.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
-version: 0.51.4
+version: 0.51.5
 keywords:
   - jaeger
   - opentracing
--- a/charts/jaeger/values.yaml
+++ b/charts/jaeger/values.yaml
@@ -551,9 +551,7 @@ esIndexCleaner:
     runAsUser: 1000
   annotations: {}
   image: jaegertracing/jaeger-es-index-cleaner
-  # tag: 1.22
   imagePullSecrets: []
-  tag: latest
   pullPolicy: Always
   cmdlineParams: {}
   extraEnv: []


pavelnikolov:
@manuelbonk why don't you want it to be set by default to `latest`? You are free to override it if this doesn't work for you.

manuelbonk:
I use [helmfile](https://github.com/roboll/helmfile) to deploy `jaeger-collector`, `jaeger-agent`, `jaeger-query` and `jaeger-es-index-cleaner` using this chart. I prefer to use the same fixed version of all of these components' images to mitigate compatibility issues. All but `jaeger-es-index-cleaner` use `jaeger.image.tag` by default (which is set to `.Chart.appVersion` by default). It doesn't make sense to me why the cronjob's version shouldn't be set to the chart's by default.

pavelnikolov:
@manuelbonk  I understand and I agree now. Make sure you bump the chart version, please.

mehta-ankit:
> I use [helmfile](https://github.com/roboll/helmfile) to deploy `jaeger-collector`, `jaeger-agent`, `jaeger-query` and `jaeger-es-index-cleaner` using this chart. I prefer to use the same fixed version of all of these components' images to mitigate compatibility issues. All but `jaeger-es-index-cleaner` use `jaeger.image.tag` by default (which is set to `.Chart.appVersion` by default). It doesn't make sense to me why the cronjob's version shouldn't be set to the chart's by default.

Why not ? Isn't index cleaner script version controlled from jaeger repo as well ?

pavelnikolov:
@mehta-ankit it seems that it is set to `latest` by default. This PR changes that.

pavelnikolov:
However, with the above being said, I wonder if it is possible to have a situation when the ES index cleaner tag doesn't get incremented. In such case it might be a good idea to leave it default to `latest` 🤔

mehta-ankit:
> However, with the above being said, I wonder if it is possible to have a situation when the ES index cleaner tag doesn't get incremented. In such case it might be a good idea to leave it default to `latest` 🤔

lol my bad 😆 . I got confused got I remembered from a previous change we had removed the key `tag`:
https://github.com/jaegertracing/helm-charts/commit/c63f2c7546f402279f17b57b837124117395a3cf#diff-67f650e4d755c78cf8961c77f7703da8dd8d2d8126ee00b0293cc1cb9e8896efR541
But looks like for esIndexCleaner `tag: latest` was left.
This LGTM

mehta-ankit:
> However, with the above being said, I wonder if it is possible to have a situation when the ES index cleaner tag doesn't get incremented. In such case it might be a good idea to leave it default to `latest` 🤔

Is that possible 🤔 .Looking at the current code in Jaeger repo, dockerfiles don't mention a tag specifically
https://github.com/jaegertracing/jaeger/blob/main/cmd/es-index-cleaner/Dockerfile
It seems like all images get releases for each component when a new Jaeger release is cut.

Can we verify this with one of the maintainers for Jaeger maybe ?

pavelnikolov:
It seems that they publish it for every release.

Screen Shot 2021-12-09 at 2 17 18 pm
Screen Shot 2021-12-09 at 2 17 52 pm

Issue #345, opened 2022-03-08
Helm Chart incorrectly has appVersion 1.31.0


mjnagel:
**Describe the bug**
https://github.com/jaegertracing/helm-charts/commit/d7671badf15944ad4721d34f663e444e42bb2511 bumped the chart version and appVersion but did not actually update to the latest CRD and image for the operator.

Going a step further, even overriding the helm values to use the 1.31.0 image will result in errors due to the introduction of cert management in https://github.com/jaegertracing/jaeger-operator/pull/1708

**To Reproduce**
1. Deploy the latest helm chart
2. Validate that the 1.30.0 image is being used despite appVersion of 1.31.0

**Expected behavior**
I would expect 1.31.0 to be deployed based on the appVersion.

hairmare:
my bad, thanks for noticing

kinda explains why i didn't run into any issues with the upgrade 😅

eanton86:
We've also run across this issue, after updating to 1.31 or 1.32 jaeger-operator fails.

sta-szek:
I faces the same. The last working version is `1.30`

mjnagel:
Issues when overriding the image are the same as reported in https://github.com/jaegertracing/helm-charts/issues/351 . There seems to be a lack of quality reviewing of updates...CRDs are rarely updated and it seems like the changes are not even tested based on the cert manager issues.

PR #350, opened 2022-04-08
jaeger-operator update to 1.32

LP0101:
#### What this PR does

Updates operator to 1.32

### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger-operator/Chart.yaml
+++ b/charts/jaeger-operator/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
 description: jaeger-operator Helm chart for Kubernetes
 name: jaeger-operator
-version: 2.29.0
-appVersion: 1.31.0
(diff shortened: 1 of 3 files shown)


cpanato (approved):
thanks!

mjnagel:
> This PR updates the operator version but doesn't update the CRD, which should be done for reference even though `--skip-crds` should be used, nor does it add the relevant resources for setting the CRD certificate. I'm not sure this was merged without these other required changes?

This was an issue with the last update as well, the image tag was not even updated properly - https://github.com/jaegertracing/helm-charts/issues/345

There's even bigger issues since 1.31.0 since the operator now requires a deployment of cert-manager to manage certs for  two webhooks - been added to the chart. I am curious if these PRs are being tested at all since it definitely appears that there have been 2 chart versions in a row that completely fail to deploy.

PR #372, opened 2022-06-07
[jaeger-operator] Refactor CRD to allow Jaeger CR creation on initial install

mjnagel:
Signed-off-by: Micah Nagel <micah.nagel@parsons.com>

#### What this PR does

https://github.com/jaegertracing/helm-charts/pull/368 moved the CRD out of the special `chart/crds` helm chart folder and into `chart/crd`. While this did result in the chart making use of the already existing `chart/templates/crd.yaml` template to install the CRD, the other result is that you must install the helm chart without `jaeger.create` set to true, then upgrade after the CRD has been installed. If we make use of the `chart/crds` folder as recommended by the first method [here](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you) then the CRDs are installed before Helm applies the other manifests.

This unknowingly introduced what I would consider a breaking change into the chart - you now are required to install the operator separately from the `Jaeger` CR (install then upgrade).

Fixes https://github.com/jaegertracing/helm-charts/issues/373

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger-operator/README.md
+++ b/charts/jaeger-operator/README.md
@@ -15,7 +15,8 @@ This chart bootstraps a jaeger-operator deployment on a [Kubernetes](http://kube
 ## Prerequisites
 
 - Kubernetes 1.19+
-- cert-manager 1.6.1+ instaled 
+- Helm 3
+- cert-manager 1.6.1+ installed, or certificate for webhook service in a secret
 
 ## Check compability matrix
 See the compatibility matrix [here](./COMPATIBILITY.md).
@@ -60,7 +61,6 @@ The following table lists the configurable parameters of the jaeger-operator cha
 | `image.pullPolicy`      | Controller container image pull policy                                                                      | `IfNotPresent`                  |
 | `jaeger.create`         | Jaeger instance will be created                                                                             | `false`                         |
 | `jaeger.spec`           | Jaeger instance specification                                                                               | `{}`                            |
-| `crd.install`           | CustomResourceDefinition will be installed                                                                  | `true`                          |
 | `rbac.create`           | All required roles and rolebindings will be created                                                         | `true`                          |
 | `serviceAccount.create` | Service account to use                                                                                      | `true`                          |
 | `rbac.pspEnabled`       | Pod security policy for pod will be created and included in rbac role                                       | `false`                         |
@@ -81,6 +81,13 @@ You can also specify any non-array parameter using the `--set key=value[,key=val
 ```console
 $ helm install jaegertracing/jaeger-operator --name my-release \
     --set rbac.create=false
+```
+
+To install the chart without creating the CRDs (any files under `chart/crds`) make use of the `--skip-crds` flag. For example,
+
+```console
+$ helm install jaegertracing/jaeger-operator --name my-release \
+    --skip-crds
 ```
 
 ## After the Helm Installation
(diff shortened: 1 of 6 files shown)


mjnagel:
The biggest downside of moving this to the `crds` folder is that Helm templating is not supported. My understanding is that the labels/annotations added in the previous PR are not functionally necessary so this is not an issue, but if those are required the issue is trickier. 

An alternative I have seen if we want the best of both worlds (helm templating in the CRD + installs with `jaeger.create`): a pre-install / pre-upgrade job that mounts a configmap with the CRDs and applies them against the cluster. Would be more of a lift and introduce a new dependency on a kubectl image, but I could modify this PR if that seems like a better route.

batazor:
@czomo Maybe we should put `crd.yaml` back in the `crds` folder?

czomo:
I completely missed that use case cause I tested it on already updated CRD/existing cluster. Thanks for pointing this out. Yes, we should move it back to `crds`

charts/jaeger-operator/crds/crd.yaml
```diff
@@ -2,10 +2,8 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: jaegers.jaegertracing.io
-  annotations:
-    cert-manager.io/inject-ca-from: {{ default {{ .Release.Namespace }} .Values.certs.certificate.namespace }}/{{ default "jaeger-operator-service-cert" .Values.certs.certificate.secretName }}
   labels:
-{{ include "jaeger-operator.labels" . | indent 4 }}
+    app: jaeger-operator
```
czomo, line 6: How about following the generic label convenction as in [here](https://github.com/jaegertracing/helm-charts/blob/main/charts/jaeger-operator/templates/_helpers.tpl#L45)

mjnagel: I can absolutely add these, I had just reverted to what was set in the past I beleive. Since we can't make use of helm templating here does it make sense to just make both labels `jaeger-operator`?
```suggestion
    app.kubernetes.io/name: jaeger-operator
    app.kubernetes.io/instance: jaeger-operator
```

czomo (approved):

charts/jaeger-operator/README.md
```diff
@@ -82,6 +81,13 @@ $ helm install jaegertracing/jaeger-operator --name my-release \
     --set rbac.create=false
 ```
 
+To install the chart without creating the CRDs (any files under `chart/crds`) make use of the `--skip-crds` flag. For example,
```
czomo, line 84: Another thing that came to my mind. As `crds` folder is helmv3 feature only then we should add note to Prerequisites section. WDYT?

mjnagel: Updated, I also clarified the cert-manager pre-requisite since ultimately the requirement is just for a certificate for the webhook.

czomo:
@mjnagel Let's squash it
@batazor Would you like to look at it as code owner?  Thanks

mjnagel:
@czomo can do, should I also update to a new patch version since https://github.com/jaegertracing/helm-charts/pull/371 was merged?

czomo:
> @czomo can do, should I also update to a new patch version since #371 was merged?

Yes, indeed. Rebase and squash is all we need :rocket:

PR #376, opened 2022-06-20
Fix certificate condition in deployment file for #375

### LINKED ISSUE #375: Disabling certs creation in v2.32.2 helm chart breaks the jaeger deployment
**Describe the bug**
v2.32.x seems to be introducing admission webhooks and associated certificates. But if I want to disable them all, its not possible to do so as per https://github.com/jaegertracing/helm-charts/blob/jaeger-operator-2.32.2/charts/jaeger-operator/templates/deployment.yaml#L52

Are they mandatory going forward ? 
Ref - https://www.jaegertracing.io/docs/1.35/operator/#prerequisite 

**To Reproduce**
1. Disable webhooks and certificates via values.yaml
```
certs:
  certificate:
    create: false
  issuer:
    create: false
webhooks:
  mutatingWebhook:
    create: false
  service:
    create: false
  validatingWebhook:
    create: false
```

2. Try to upgrade jaeger helm chart from 2.27.0 to 2.32.2

3. Pod will stay in Waiting/ContainerCreating state forever and will show below message if we do kubectl describe on pod,

```
  Warning  FailedMount  50s (x10 over 5m)    kubelet            MountVolume.SetUp failed for volume "cert" : secret "jaeger-operator-service-cert" not found
  Warning  FailedMount  42s (x2 over 2m57s)  kubelet            Unable to attach or mount volumes: unmounted volumes=[cert], unattached volumes=[cert kube-api-access-7bkf8]: timed out waiting for the condition

```
**Expected behavior**
Currently deployment.yaml tries to load volume from certificate secret even if we disable cert creation. It needs to have the condition to check for that flag before volume/volume mount creation. That way, webhooks/certificates creation won't be a mandatory configuration.

**Screenshots**

**Version (please complete the following information):**
 - OS: Linux Ubuntu
 - Jaeger version: 1.34.1
 - Deployment: Kubernetes v1.22

**What troubleshooting steps did you try?**
Try to follow https://www.jaegertracing.io/docs/latest/troubleshooting/ and describe how far you were able to progress and/or which steps did not work.

**Additional context**

AshutoshNirkhe:

- fixes #375 

#### Checklist

- [ ] Chart Version bumped


czomo:
Did you tried to deploy it with .certs.certificate.create to `off`? I guess it will end up with sth like `/tmp/k8s-webhook-server/serving-certs/tls.crt: no such file or directoryManager exited non-zero` because jaeger operator expect cert under that path.
cert-manager [pre-requisite](https://www.jaegertracing.io/docs/1.35/operator/#prerequisite) is not without a reason because it is responsible for handling certs for secure connection. You can create certs for your own however they are mandatory to work.

charts/jaeger-operator/templates/deployment.yaml
```diff
@@ -48,10 +48,12 @@ spec:
           - containerPort: 9443
             name: webhook-server
             protocol: TCP  
+          {{- if .Values.certs.certificate.create }}
```
czomo, line 51: Check out the https://github.com/jaegertracing/helm-charts/tree/main/charts/jaeger-operator#prerequisites for the operator chart and my comment in conversation. In case you don't want to use cert-manager you should create certificate for webhook service in a secret instead.

AshutoshNirkhe: @czomo thanks for the explanation. I have already configured it to use webhooks and self-signed issuer/certs now so am good. 
Basically it started with if we can disable both webhooks and hence certificates for webhooks. 
Like what if I disable both the webhooks ? (https://github.com/jaegertracing/helm-charts/blob/jaeger-operator-2.32.2/charts/jaeger-operator/values.yaml#L22)
Why should I still care about certificates ?

czomo: jaeger-operator can be deployed in [multiple ways](https://www.jaegertracing.io/docs/1.24/operator/#install-modes), that's why there is option to disable specific components.

webhooks in general are integrated part of jaeger because they are responsible for validating jeager resources(validating webhook) or in case of mutating webhooks adding jaeger sidecars. However there are some cases when you want do disable them because of  conflicts with other resources.  Be carefull about that.
One of the purpose of certs is handling secure connection between service and webhooks, more features will be added in the future.

AshutoshNirkhe: Ok so I think its wiser to not disable any of that :) We don't this PR anymore in that case.

avishefi: We should be able to disable creating the certificates so people can use their own `cert-manager` issuers and settings. Is this PR being merged?

em-le-ts:
Is any update for this PR? We're facing the same issue. Many thanks

Issue #104, opened 2020-05-14
Jaeger Chart :- need custom labels support in query-ing.yaml


rgssingh:
Hi Team,

Can you please provide the custom labels support in query-ing.yaml. 

For example :-

{{- toYaml .Values.query.ingress.labels | nindent 4 }}

Thanks

naseemkullah:
Hi @rgssingh, would you be willing to open a PR to add this functionality?
Also I am curious what is your use case if you don't mind sharing?

rgssingh:
Hi @naseemkullah,

I would be happy to do that but I need to go through approval process with my organization before contributing with a PR and that would take some time.

It would be good if you can put this for me.

Also regarding use case : We are using Traefik with custom label selector and this Ingress is not getting picked up by Traefik due to missing label that we are attaching to all our Ingress Rules.

NicolasLemince:
Hello @naseemkullah, 
Is there any step forward regarding this request ? I need this feature and it doesn't seem to be available. If not I can do a PR.

naseemkullah:
Hi @NicolasLemince, a PR sounds good.

PR #377, opened 2022-06-20
Adding custom labels in query and collector ingresses

NicolasLemince:
Signed-off-by: Nicolas <nicorabatel@gmail.com>

#### What this PR does
Adding values to be able to add custom labels in query and collector ingresses

- fixes https://github.com/jaegertracing/helm-charts/issues/104

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.30.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
-version: 0.56.6
+version: 0.56.7
 keywords:
   - jaeger
   - opentracing
--- a/charts/jaeger/README.md
+++ b/charts/jaeger/README.md
@@ -336,6 +336,8 @@ If extra protection of the Jaeger UI is needed, then the oAuth2 sidecar can be e
 sidecar acts as a security proxy in front of the Jaeger Query service and enforces user authentication before reaching
 the Jaeger UI. This method can work with any valid provider including Keycloak, Azure, Google, GitHub, and more.
 
+Offical docs [here](https://oauth2-proxy.github.io/oauth2-proxy/docs/behaviour)
+
 Content of the `jaeger-values.yaml` file:
 
 ```YAML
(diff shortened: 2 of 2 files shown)


charts/jaeger/templates/collector-ing.yaml
```diff
@@ -7,7 +7,7 @@ kind: Ingress
 metadata:
   name: {{ template "jaeger.collector.name" . }}
   labels:
-    {{- include "jaeger.labels" . | nindent 4 }}
+    {{- toYaml .Values.collector.ingress.labels | nindent 4 }}
```
naseemkullah, line 10:
```suggestion
    {{- include "jaeger.labels" . | nindent 4 }}
    {{- if .Values.query.ingress.labels }}
    {{- toYaml .Values.collector.ingress.labels | nindent 4 }}
    {{- end }}
```

what about something like this that keeps the common labels, but allows folks to add additional labels to the ingress?

charts/jaeger/templates/query-ing.yaml
```diff
@@ -7,7 +7,7 @@ kind: Ingress
 metadata:
   name: {{ template "jaeger.query.name" . }}
   labels:
-    {{- include "jaeger.labels" . | nindent 4 }}
+    {{- toYaml .Values.query.ingress.labels | nindent 4 }}
```
naseemkullah, line 10: same as above

naseemkullah (changes requested):
I propose that rather than replace the labels, add an option to append labels - would that work for you?

charts/jaeger/templates/collector-ing.yaml
NicolasLemince: Totally agree, this is done.

charts/jaeger/templates/query-ing.yaml
NicolasLemince: Totally agree, this is done.

NicolasLemince:
That's ok, I made the changes + I signed the last commit

naseemkullah (approved):

PR #394, opened 2022-08-18
support allInOne otlp grpc and http

dorsegal:

- fixes #

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/charts/jaeger/templates/allinone-collector-svc.yaml
+++ b/charts/jaeger/templates/allinone-collector-svc.yaml
@@ -21,6 +21,22 @@ spec:
     - name: http-c-binary-trft
       port: 14268
       targetPort: 0
+    {{- if or .Values.allInOne.service.otlp.grpc .Values.allInOne.service.otlp.http }}
+    - name: otlp-grpc
+      port: {{ default 4317 .Values.allInOne.service.otlp.grpc.port }}
+    {{- if and (eq .Values.allInOne.service.type "NodePort") (.Values.allInOne.service.otlp.grpc.nodePort) }}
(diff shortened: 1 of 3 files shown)


charts/jaeger/values.yaml
```diff
@@ -16,7 +16,7 @@ fullnameOverride: ""
 allInOne:
   enabled: false
   image: jaegertracing/all-in-one
-  tag: 1.29.0
+  tag: 1.37.0
```
sparty02, line 19: Can we just remove this instead?  That will be consistent with the rest of the chart, which defaults image tags to the chart's appVersion

sparty02 (changes requested):

charts/jaeger/values.yaml
jplayout: Why not using the tag "latest" ?

charts/jaeger/templates/allinone-collector-svc.yaml
```diff
@@ -21,6 +21,22 @@ spec:
     - name: http-c-binary-trft
       port: 14268
       targetPort: 0
+    {{- if or .Values.allInOne.service.otlp.grpc .Values.allInOne.service.otlp.http }}
```
naseemkullah, line 24: so if only one of grpc or http are defined, both are set?

naseemkullah:
Thanks @dorsegal , https://github.com/jaegertracing/helm-charts/pull/399/files was merged a while back. Does that fulfill your use case?

dorsegal:
> 

It looks like it almost the same PR. LGTM.

PR #396, opened 2022-08-22
[jaeger] Add ttlSecondsAfterFinished entry to esIndexCleaner job

### LINKED ISSUE #393: [Feature]: TTL setting for esIndexCleaner jobs after completion
### Requirement

As a Kubernetes Administrator
I would like to set ttlSecondsAfterFinished for the es index cleaner
so that old jobs in completed state get removed sooner

### Problem

image

Old jobs in completet state remain for hours

### Proposal

Like in esRollover i would like ttlSecondsAfterFinished entry for esIndexCleaner
https://github.com/jaegertracing/helm-charts/blob/eff00f612e3acf0cb335b169c6186ff581a40d2b/charts/jaeger/values.yaml#L692

### Open questions

_No response_

Mahagon:
Add ttlSecondsAfterFinished entry to esIndexCleaner for removing completed jobs of the ElasticSearch Index Cleaner after the configured amount of time

#### What this PR does

Make ttlSecondsAfterFinished for esIndexCleaner configurable

- fixes #393

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/templates/es-index-cleaner-cronjob.yaml
+++ b/charts/jaeger/templates/es-index-cleaner-cronjob.yaml
@@ -21,6 +21,9 @@ spec:
       {{- if .Values.esIndexCleaner.activeDeadlineSeconds }}
       activeDeadlineSeconds: {{ .Values.esIndexCleaner.activeDeadlineSeconds }}
       {{- end }}
+      {{- if .Values.esIndexCleaner.ttlSecondsAfterFinished }}
+      ttlSecondsAfterFinished: {{ .Values.esIndexCleaner.ttlSecondsAfterFinished }}
+      {{- end }}
       template:
         metadata:
           {{- if .Values.esIndexCleaner.podAnnotations }}
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.37.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
-version: 0.59.0
+version: 0.60.0
 keywords:
   - jaeger
   - opentracing
--- a/charts/jaeger/templates/es-lookback-cronjob.yaml
+++ b/charts/jaeger/templates/es-lookback-cronjob.yaml
@@ -21,6 +21,9 @@ spec:
       {{- if .Values.esLookback.activeDeadlineSeconds }}
       activeDeadlineSeconds: {{ .Values.esLookback.activeDeadlineSeconds }}
       {{- end }}
+      {{- if .Values.esLookback.ttlSecondsAfterFinished }}
+      ttlSecondsAfterFinished: {{ .Values.esLookback.ttlSecondsAfterFinished }}
+      {{- end }}
       template:
         metadata:
           {{- if .Values.esLookback.podAnnotations }}
--- a/charts/jaeger/templates/es-rollover-cronjob.yaml
+++ b/charts/jaeger/templates/es-rollover-cronjob.yaml
@@ -21,6 +21,9 @@ spec:
       {{- if .Values.esRollover.activeDeadlineSeconds }}
       activeDeadlineSeconds: {{ .Values.esRollover.activeDeadlineSeconds }}
       {{- end }}
+      {{- if .Values.esRollover.ttlSecondsAfterFinished }}
+      ttlSecondsAfterFinished: {{ .Values.esRollover.ttlSecondsAfterFinished }}
+      {{- end }}
       template:
         metadata:
           {{- if .Values.esRollover.podAnnotations }}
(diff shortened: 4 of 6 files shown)


charts/jaeger/templates/es-index-cleaner-cronjob.yaml
```diff
@@ -21,6 +21,9 @@ spec:
       {{- if .Values.esIndexCleaner.activeDeadlineSeconds }}
       activeDeadlineSeconds: {{ .Values.esIndexCleaner.activeDeadlineSeconds }}
       {{- end }}
+      {{- with .Values.esIndexCleaner.ttlSecondsAfterFinished }}
```
mehta-ankit, line 24: I see this feature is stable in 1.23 version of K8s. What about 1.21 or 1.19. Do you know if it would work for folks using that version of k8s or would we have to use bitnami common chart capabilities to make sure its backwards compatible.

mehta-ankit (commented):
left few comments.

mehta-ankit:
@Mahagon You also need to update the chart version.

charts/jaeger/templates/es-index-cleaner-cronjob.yaml
```diff
@@ -21,6 +21,9 @@ spec:
...
+      ttlSecondsAfterFinished: {{ . }}
+      {{- end }}
```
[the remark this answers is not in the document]
mehta-ankit: Also [rollover cronjob](https://github.com/jaegertracing/helm-charts/blob/main/charts/jaeger/templates/es-rollover-cronjob.yaml#L20) does not use it for the job it creates.
If we are adding it we should add it to all the cronjob's we have I think.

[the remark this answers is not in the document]
Mahagon: true

did the same for the other cronjobs

charts/jaeger/templates/es-index-cleaner-cronjob.yaml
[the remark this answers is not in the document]
Mahagon: `successfulJobsHistoryLimit` and `failedJobsHistoryLimit` is already implemented and defaults to 3.

i just tested it running a cronjob multiple times with `ttlSecondsAfterFinished` of 5 minutes and `successfulJobsHistoryLimit` = 2

running the job 3 times within the 5 minutes keeps 2 jobs
those get deleted after 5 minutes

if i run the job once it just gets deleted after 5 minutes

so the settings seem to work fine together

Mahagon: Yes it's a feature of [1.23](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/) 

But i'm unsure how to make it backwards compatible. I never had to do this before.

mehta-ankit: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
Feature | Default | Stage | Since | Until
-- | -- | -- | -- | --
TTLAfterFinished | false | Alpha | 1.12 | 1.20
TTLAfterFinished | true | Beta | 1.21 | 1.22
TTLAfterFinished | true | GA | 1.23 | -

Found this in the K8s docs. It looks like it was beta in 1.21, so should work (only if beta features have been enabled on your K8s clusters).

@naseemkullah thoughts ?

mehta-ankit: @Mahagon what we can also do is, only add the key in the job spec, when the value is defined in the values.yaml file. So instead of 120seconds, we keep it commented or empty value and check for it. so people who have support for it can add it and other do not have to add it to their values.yaml. Thoughts ?

Mahagon: i would be fine with that too :)

mehta-ankit: Thanks for that.

mehta-ankit (approved):

mehta-ankit:
@Mahagon thanks for all the changes. 🙏🏽 
Are you comfortable squashing all your commits into 1 or 2 commits ? That would be great.
Also I don't think the commits are GPG signed. Can you do that too plz.

Mahagon:
Sure :) squashed the commits and added my gpg signature

mehta-ankit:
@Mahagon i merged another PR recently, so you'd have to bump the chart version.

mehta-ankit (approved):
LGTM

charts/jaeger/templates/es-index-cleaner-cronjob.yaml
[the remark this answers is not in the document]
mehta-ankit: thank you

PR #439, opened 2023-01-30
Add support for configurable hpa

DrFaust92:
same as dynamic ingess version. this will allow k8s 1.25/1.26 deployments

### DIFF
--- a/charts/jaeger/templates/ingester-hpa.yaml
+++ b/charts/jaeger/templates/ingester-hpa.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.ingester.autoscaling.enabled }}
-apiVersion: autoscaling/v2beta1
+apiVersion: {{ include "common.capabilities.hpa.apiVersion" ( dict "context" $ ) }}
 kind: HorizontalPodAutoscaler
 metadata:
   name: {{ template "jaeger.ingester.name" . }}
@@ -8,7 +8,7 @@ metadata:
     app.kubernetes.io/component: ingester
 spec:
   scaleTargetRef:
-    apiVersion: apps/v1
+    apiVersion: {{ include "common.capabilities.deployment.apiVersion" . }}
     kind: Deployment
     name: {{ template "jaeger.ingester.name" . }}
   minReplicas: {{ .Values.ingester.autoscaling.minReplicas }}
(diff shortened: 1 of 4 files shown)


charts/jaeger/templates/_capabilities.tpl
```diff
@@ -0,0 +1,154 @@
+{{/* vim: set filetype=mustache: */}}
+
+{{/*
+Return the target Kubernetes version
+*/}}
```
mehta-ankit, line 5: @DrFaust92 i dnt think we need these here. We already include common bitnami chart, which has all these defined: https://github.com/bitnami/charts/blob/main/bitnami/common/templates/_capabilities.tpl

mehta-ankit (changes requested):

charts/jaeger/templates/_capabilities.tpl
mehta-ankit: your change in hpa yaml files would just work (https://github.com/bitnami/charts/blob/main/bitnami/common/templates/_capabilities.tpl#L132)

DrFaust92: `{{- if semverCompare "<1.23-0" (include "common.capabilities.kubeVersion" .context) -}} ` doesnt work when i test locally. 
i think it needs to be 
`{{- if semverCompare "<1.23-0" (include "common.capabilities.kubeVersion" .) -}}` instead
otherwise this errors on:
```
Error: template: jaeger/templates/ingester-hpa.yaml:2:15: executing "jaeger/templates/ingester-hpa.yaml" at <include "common.capabilities.hpa.apiVersion" $>: error calling include: template: jaeger/charts/common/templates/_capabilities.tpl:133:7: executing "common.capabilities.hpa.apiVersion" at <semverCompare "<1.23-0" (include "common.capabilities.kubeVersion" .context)>: error calling semverCompare: Invalid Semantic Version

Use --debug flag to render out invalid YAML
```


i think i need to tweak the HPA resource a bit more, need to change `targetAverageUtilization` to `averageUtilization` with some conditional on api version

DrFaust92: nvm, found the issue and remove the capabilities file

mehta-ankit (commented):
This LGTM.
While we are doing this, can you please enhance the HPA to also allow users to add scalingBehavior to their HPA's.
Similar to this: https://github.com/cortexproject/cortex-helm-chart/blob/master/templates/distributor/distributor-hpa.yaml#L35

mehta-ankit:
Once you do that, a chart version bump, squashing of the commits and GPG signing the commit would be few other things remaining. Thanks for the contribution @DrFaust92 🙏🏼 , appreciate it.

DrFaust92:
[mehta-ankit](https://github.com/mehta-ankit) sure np will do those

charts/jaeger/templates/ingester-hpa.yaml
```diff
@@ -18,10 +18,26 @@ spec:
...
         name: cpu
+        {{- if eq "autoscaling/v2" (include "common.capabilities.hpa.apiVersion" ( dict "context" $ )) }}        
+        target:
+          averageUtilization: {{ .Values.ingester.autoscaling.targetCPUUtilizationPercentage | default 80 }}
+          type: Utilization
+        {{- else }}    
         targetAverageUtilization: {{ .Values.ingester.autoscaling.targetCPUUtilizationPercentage | default 80 }}
+        {{- end }}
+  {{- with .Values.ingester.autoscaling.behavior }}
+  behavior:
```
mehta-ankit, line 40: let's add default value as {} in values.yaml file also.

DrFaust92: done

mehta-ankit (approved):
LGTM ✅

PR #478, opened 2023-06-12
[jaeger] Add configurable OLTP port names in jaeger collector services

cnvergence:
#### What this PR does

Adding a prefix to OLTP service port names, resolves issues with istio.
As per https://github.com/jaegertracing/helm-charts/issues/344#issuecomment-1583708032 it can unblock traces for oltp services.
```bash

Info [IST0118] (Service jaeger-collector.test) Port name oltp-http (port: 4318, targetPort: 0) doesn't follow the naming convention of Istio port.
Info [IST0118] (Service jaeger-collector.test) Port name oltp-grpc (port: 4317, targetPort: 0) doesn't follow the naming convention of Istio port.
```
Potentially possible problems similar to this old [issue](https://github.com/jaegertracing/jaeger-operator/issues/907).
#### Which issue this PR fixes

*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format,
will close that issue when PR gets merged)*

- https://github.com/jaegertracing/helm-charts/issues/344#issuecomment-1100166786 (partially)

#### Checklist

- [x] [DCO](https://github.com/jaegertracing/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [x] Commits are [GPG signed](https://docs.github.com/en/github/authenticating-to-github/about-commit-signature-verification)
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (`[jaeger]` or `[jaeger-operator]`)
- [x] README.md has been updated to match version/contain new values

Issue #344, opened 2022-03-01
Add option to change port names


KarstenSiemer:
## Requirement - what kind of business use case are you trying to solve?

Hi there!
We are currently using Jaeger and want to add authentication and authorization to the UI.
We are doing this via istio service mesh and a dex/oauth2-proxy flow.

## Problem - what in Jaeger blocks you from solving the requirement?

Sadly though, istio does not recognize the jaeger-query port as http and we need to prefix the 
port name with `http` to force the selection to have rules for the traffic to apply.
The helm chart jaeger however, doesn't give an option to configure port names.

When some other protocol than http get's selected for the port, the http rules we supply
to activate the authentication flow never apply.

As a quick fix we deploy an additional service with the same selector to force selection of http, but
it would be nice if we could just configure the port name.

## Proposal - what do you suggest to solve the problem or improve the existing situation?
```
query:
  service:
    annotations: {}
    type: ClusterIP
    # List of IP ranges that are allowed to access the load balancer (if supported)
    loadBalancerSourceRanges: []
    port: 80
    # Specify a custom port name
    portName: query
    # Specify a custom target port (e.g. port of auth proxy)
    # targetPort: 8080
    # Specify a specific node port when type is NodePort
    # nodePort: 32500
```
I'd like to give the service sections an interface to configure the port names via a `name` field.
The default value should be is currently hardcoded. Folks that use istio or something else can then just insert whatever they desire. Like `query.service.portName: http-query`

For us doing it on the query service would suffice, but we should follow through and make the name configurable on all services.
e.g agent:

```
  service:
    annotations: {}
    # List of IP ranges that are allowed to access the load balancer (if supported)
    loadBalancerSourceRanges: []
    type: ClusterIP
    # zipkinThriftPort :accept zipkin.thrift over compact thrift protocol
    zipkinThrift:
      name: zipkin-compact
      port: 5775
    # compactPort: accept jaeger.thrift over compact thrift protocol
    compact:
      name: jaeger-compact
      port: 6831
    # binaryPort: accept jaeger.thrift over binary thrift protocol
    binary:
      name: jaeger-binary
      port: 6832
    # samplingPort: (HTTP) serve configs, sampling strategies
    sampling:
      name: http
      port: 5778
```

## Any open questions to address

* We could also make this easier to configure by adding an istio options field and activate port naming convention and auto configure correct prefixes in this chart.
[...]

PR #479, opened 2023-06-23
[jaeger] Fix otlp service regression in 0.71.5

spadger:
Version 0.71.5 added configurable otlp port names to the collector service, but introduced some typos, which break otlp for http & gRPC. I'm not sure why they would be configurable, so I haven't also made the port names on the actual deployment configurable to match the service, but at least this should fix connectivity
Signed-off-by: Jon Bates <jonathan.bates@team.bumble.com>

#### What this PR does
Fixes regression in 0.71.5

Can't see one that has been raised

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/values.yaml
+++ b/charts/jaeger/values.yaml
@@ -36,9 +36,9 @@ allInOne:
     collector:
       otlp:
         grpc:
-          name: otlp-grpc
+          name: grpc-otlp
         http:
-          name: otlp-http
+          name: http-otlp
   ingress:
     enabled: false
     # For Kubernetes >= 1.18 you should specify the ingress-controller via the field ingressClassName
@@ -379,11 +379,11 @@ collector:
       # nodePort:
     otlp:
       grpc:
-        name: otlp-grpc
+        name: grpc-otlp
         # port: 4317
         # nodePort:
       http:
-        name: otlp-grpc
+        name: http-otlp
         # port: 4318
         # nodePort:
   ingress:
--- a/charts/jaeger/values.yaml
+++ b/charts/jaeger/values.yaml
@@ -36,9 +36,9 @@ allInOne:
     collector:
       otlp:
         grpc:
-          name: gtpc-otlp
+          name: otlp-grpc
         http:
-          name: http-otlp
+          name: otlp-http
   ingress:
     enabled: false
     # For Kubernetes >= 1.18 you should specify the ingress-controller via the field ingressClassName
@@ -379,11 +379,11 @@ collector:
       # nodePort:
     otlp:
       grpc:
-        name: grpc-otlp
+        name: otlp-grpc
         # port: 4317
         # nodePort:
       http:
-        name: http-otlp
+        name: otlp-grpc
         # port: 4318
         # nodePort:
   ingress:
--- a/charts/jaeger/values.yaml
+++ b/charts/jaeger/values.yaml
@@ -36,9 +36,9 @@ allInOne:
     collector:
       otlp:
         grpc:
-          name: grpc-otlp
+          name: otlp-grpc
         http:
-          name: http-otlp
+          name: otlp-http
   ingress:
     enabled: false
     # For Kubernetes >= 1.18 you should specify the ingress-controller via the field ingressClassName
@@ -379,11 +379,11 @@ collector:
       # nodePort:
     otlp:
       grpc:
-        name: grpc-otlp
+        name: otlp-grpc
         # port: 4317
         # nodePort:
       http:
-        name: http-otlp
+        name: otlp-http
         # port: 4318
         # nodePort:
   ingress:
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.45.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
-version: 0.71.5
+version: 0.71.6
 # CronJobs require v1.21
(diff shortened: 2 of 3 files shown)


charts/jaeger/values.yaml
```diff
@@ -379,11 +379,11 @@ collector:
       # nodePort:
     otlp:
       grpc:
-        name: grpc-otlp
+        name: otlp-grpc
         # port: 4317
         # nodePort:
       http:
-        name: http-otlp
+        name: otlp-grpc
```
mehta-ankit, line 386:
```suggestion
        name: otlp-http
```

mehta-ankit:
@spadger Thanks for catching this. It was my bad, I missed this in the previous MR.

mehta-ankit (commented):
We should also make deployment port names configurable.
https://github.com/cnvergence/jaeger-helm-charts/blob/main/charts/jaeger/templates/collector-deploy.yaml#L122

spadger:
> @spadger Thanks for catching this. It was my bad, I missed this in the previous MR.

Welcome.  I just saw the linter output - looks like I might not have done any better
```
Error: Service "jaeger-flmntaf7jw-collector" is invalid: spec.ports[3].name: Duplicate value: "otlp-grpc"
```

spadger:
@mehta-ankit I can make them configurable, but could you help understand why we'd do this?  As far as I understand, port `name`s are just arbitrary tokens so that k8s uses to match a service's port with a pod's port. A hard-coded `otlp-grpc` is descriptive, so could you explain the case where someone would change this?

mehta-ankit:
> @mehta-ankit I can make them configurable, but could you help understand why we'd do this? As far as I understand, port `name`s are just arbitrary tokens that k8s uses to match a service's port with a pod's port. A hard-coded `otlp-grpc` is descriptive, so could you explain the case where someone would change this?

People had issues with certain port names because of Istio. I don't have all the details but this should explain: https://github.com/jaegertracing/helm-charts/issues/344#issuecomment-1583708032

I agree with you, but keeping them configurable and making the OG values the default, its a win-win for us who dnt need to change port names and for others who are having issues. Do you agree ?

charts/jaeger/values.yaml
spadger: You're right that this was a mistake, but the actual format should have been `http-otlp`.  Otherwise, we'd have caused a regression you attempted to fix in `0.71.5`!

mehta-ankit:
> OK, that's complete, and I found my own typo that caused the lining to fail

@spadger i am confused. Why are we not changing the values.yaml to be what it was before 0.71.5 change:
`otlp-grpc` & `otlp-http` ??
Look at this commit: https://github.com/jaegertracing/helm-charts/pull/478/files#diff-30fbc05152b26afaa2337764ff007b0240c79e25d76fb7ec985a18adc0ee3140

spadger:
> @spadger i am confused. Why are we not changing the values.yaml to be what it was before 0.71.5 change: `otlp-grpc` & `otlp-http` ?? Look at this commit: https://github.com/jaegertracing/helm-charts/pull/478/files#diff-30fbc05152b26afaa2337764ff007b0240c79e25d76fb7ec985a18adc0ee3140

The istio bug you pointed at was caused because the port name needed to start with `http-`, and I saw [here](https://github.com/jaegertracing/helm-charts/pull/478/files#diff-67f650e4d755c78cf8961c77f7703da8dd8d2d8126ee00b0293cc1cb9e8896efR41) that you even made that change to `values.yaml`.  I figured it'd be better to keep the port names named like that by default because:
* Most people won't care
* The istio crowd have a workaround

If you prefer, I'm happy to change the port names to `otlp-grpc` & `otlp-http` , and let the istio folks override config (I'm not fussed either way)

mehta-ankit:
@cnvergence FYI, https://github.com/jaegertracing/helm-charts/pull/479#issuecomment-1604489712
Since you made the previous change, I wanted to let you know that you would want to change values for your helm chart deployment to match whatever you need. In this PR we are making those values what they were before your change so as to not break backwards compatibility.

mehta-ankit (approved):
LGTM

cnvergence:
Oh, nice catch, thanks @spadger @mehta-ankit 
and I fully agree that the default names should have old values.
I completely missed that when I was refactoring my PR, sorry about that

spadger:
Hey @mehta-ankit 
So I happen to be using jaeger in an istio environment.  Nothing as special like the bug you linked to; simply "We have an isito gateway, and want to report otlp spans".  Because the port doesn't start with `grpc-`, Istio sends the spans as an http POST instead of a gRPC call.  Do you reckon its worth switching the ports to `grpc-otlp` and `http-otlp` to fix this papercut?  I was really lucky because after a few hours of head-banging, I remembered the port-naming issue, but others may not be so lucky

mehta-ankit:
> Hey @mehta-ankit So I happen to be using jaeger in an istio environment. Nothing as special like the bug you linked to; simply "We have an isito gateway, and want to report otlp spans". Because the port doesn't start with `grpc-`, Istio sends the spans as an http POST instead of a gRPC call. Do you reckon its worth switching the ports to `grpc-otlp` and `http-otlp` to fix this papercut? I was really lucky because after a few hours of head-banging, I remembered the port-naming chart feature, but others may not be so lucky

Hi @spadger 
I don't think using Istio is such a common thing where we should change default values which could affect anyone else depending on older names for whatever reason. 
Do you think its a better idea maybe to add some docs around Istio and port name oddity ?
https://github.com/jaegertracing/helm-charts/tree/main/charts/jaeger

PR #486, opened 2023-07-06
[jaeger operator] Fix Parameters table on README.md

mythrilium:
#### What this PR does

Fixes table formating for the README.md file
---

## Previous table visualization

image

## After fixes

image


batazor:
@mythrilium I think we should make the change in the standard way, by bumping the chart version and signing the commit

PR #494, opened 2023-08-09
[jaeger] make replicas a Helm value instead of hardcoding it

### LINKED ISSUE #493: [Feature]: make replicas spec a Helm variable rather than hardcoding to `replicas: 1`
### Requirement

We'd like to be able to specify the number of replicas using Helm.

### Problem

The replicas count is hardcoded to 1.

### Proposal

Make the replicas spec a Helm variable.

### Open questions

_No response_

lisenet:
#### What this PR does

Adds a Helm value for `allInOne.replicas`.

- fixes #493

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.45.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
-version: 0.71.10
+version: 0.71.11
 # CronJobs require v1.21
 kubeVersion: '>= 1.21-0'
 keywords:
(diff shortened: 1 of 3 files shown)


mehta-ankit (approved):

pavelnikolov:
@mehta-ankit does it make sense to have more than one replica of the all-in-one container? I think it doesn't make sense and this PR shouldn't have been merged.

mehta-ankit:
> @mehta-ankit does it make sense to have more than one replica of the all-in-one container? I think it doesn't make sense and this PR shouldn't have been merged.

you do bring up a good point and I thought about it, but since this was a backwards compatible change I went ahead and merged it.

lisenet:
@pavelnikolov our use case is that we need to deploy it with **zero** replicas so that no deployment is running but the config is there, and we flip the switch when we run a performance test to gather Istio tracing information. It is currently not possible to do so using the current chart version.

pavelnikolov:
fyi https://github.com/jaegertracing/helm-charts/pull/499

PR #499, opened 2023-08-22
[Jaeger] fix .allinone.replicas helm value

Mihail-blip:
#### When the below if statement is computed for `replicas: 0`, the result is `false` and as a consequence the deployment manifest doesn't contain `replicas` field, which in turn forces it to use the default value, which is `1`.
```yaml
spec:
  {{- if .Values.allInOne.replicas }}
  replicas: {{ .Values.allInOne.replicas }}
  {{- end }}
```

- fixes # Rather than checking value of `replicas`, we can check if `.allInOne.replicas` key has been declared. This would create a deployment with the specified number of replicas in the values file.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.45.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
-version: 0.71.13
+version: 0.71.14
 # CronJobs require v1.21
 kubeVersion: '>= 1.21-0'
 keywords:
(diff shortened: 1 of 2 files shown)


pavelnikolov:
@Mihail-blip not all commit have been signed. Could you, please, squash all the two commits into one and sign them?

Mihail-blip:
One of my first contributions, I accidentally closed the PR. 
Think I managed to gpg sign the commits, please review

pavelnikolov:
The PR is still not DCO signed. Thank you for reporting.
Fixed in https://github.com/jaegertracing/helm-charts/commit/28ab98bfb2b2c5ab4f024b7337c3bb76871a0768

Issue #443, opened 2023-02-01
[Bug]:"Failed to init storage factory","error":"failed to create primary Elasticsearch client: health check timeout: no Elasticsearch node available"


navin-rai:
### What happened?

I am using AWS Elasticsearch and trying to use it in jaeger, I have set the endpoints as per documentation, I am using latest version of jaeger helm chart apiVersion: v2, appVersion: 1.39.0, below is the config I am using for elasticsearch
 elasticsearch:
    scheme: https
    host: search-*********************.us-east-1.es.amazonaws.com
    port: 443
    user: elastic
    usePassword: true
    password: *********

### Steps to reproduce

Add AWS ES endpoints
helm install jaeger

### Expected behavior

Jaeger Collector and Jaeger Query should deploy properly on Kubernetes.

### Relevant log output

```shell
2023/02/01 13:18:20 maxprocs: Leaving GOMAXPROCS=24: CPU quota undefined
{"level":"info","ts":1675257500.4474466,"caller":"flags/service.go:119","msg":"Mounting metrics handler on admin server","route":"/metrics"}
{"level":"info","ts":1675257500.447506,"caller":"flags/service.go:125","msg":"Mounting expvar handler on admin server","route":"/debug/vars"}
{"level":"info","ts":1675257500.4477003,"caller":"flags/admin.go:129","msg":"Mounting health check on admin server","route":"/"}
{"level":"info","ts":1675257500.4477556,"caller":"flags/admin.go:143","msg":"Starting admin HTTP server","http-addr":":14269"}
{"level":"info","ts":1675257500.447806,"caller":"flags/admin.go:121","msg":"Admin server started","http.host-port":"[::]:14269","health-status":"unavailable"}
[...]

PR #514, opened 2023-10-31
[jaeger] more flexible TLS variables

### LINKED ISSUE #513: [Bug]: tls.enabled value is bound with hardcoded values
### What happened?

Hopefully this isn't a "No one gets TLS".

I just wanted to deploy jaeger with an elasticsearch statefulset deployed via the ECK operator. The ECK operator rolls out its own certificates, which is fine. 

When I activate TLS via the normal helm chart (`tls.enable` is set to true), the ES_TLS_CA variable is set to a default, hardcoded value, which is per default pumped into the environment of all relevant containers (`collector` and `query`). A secret with a specific name must be created, which contains the `ca-cert.pem` file. This already makes the chart **non-deployable**, without creating said secret, which doesn't seem directly possible via the ECK operator. As certificates are rotated via ECK, creating a manual secret by collecting the needed data seemed counter-intuitive.

Given that the elasticsearch deployment doesn't create one single secret with the expected value with certificate chain (`/es-tls/ca-cert.pem`), I found no simple way to mount the the certs and key created via the Elasticsearch deployment. An `elastisearch` resource creates a secret containing 3 values, ca.crt, tls.key and tls.crt. That secret cannot be mounted via the current chart in a way, which wouldn't block the deployemt of jaeger. 

Using `additionalSecretMounts` doesn't work, as we must activate tls, to be able to be able to use the `es.tls.ca`, `es.tls.cert` and `es.tls.key` flags.  

### Steps to reproduce

First create a simple `elasticsearch` resource via the ECK operator:

```yaml
apiVersion: elasticsearch.k8s.elastic.co/v1
kind: Elasticsearch
metadata:
  name: quickstart
spec:
    count: 1
    name: default
  version: 7.17.13
```

The deploy jaeger with the following values:

```yaml
provisionDataStore:
  cassandra: false
  elasticsearch: false

storage:
  type: elasticsearch
  elasticsearch:
    port: 9200
    scheme: https
    user: elastic
    indexPrefix: test
    host: quickstart-es-internal-http.tracing.svc
    usePassword: false
    tls:
      enabled: true
      secretName: quickstart-es-http-certs-internal

agent:
  enabled: true

collector:
  enabled: true
  extraEnv:
    - name: ES_PASSWORD
      valueFrom:
        secretKeyRef:
          name: quickstart-es-elastic-user
          key: elastic
query:
  enabled: true
  extraEnv:
    - name: ES_PASSWORD
      valueFrom:
        secretKeyRef:
          name: quickstart-es-elastic-user
          key: elastic
```

The collector and query containers cannot start, as they cannot find the expected default secret in the default directory, which is defined via the `ES_TLS_CA` via the `elasticsearch.env` helper function. The secrets are mounted per default in all related containers via:

```yaml
              - name: {{ .Values.storage.elasticsearch.tls.secretName }}
                mountPath: "/es-tls/ca-cert.pem"
                subPath: "ca-cert.pem"
```

But the secret created via the ECK Operator has the following syntax:

```yaml
ca.crt: |
  -----BEGIN CERTIFICATE-----
  xxxx
  -----END CERTIFICATE-----
tls.crt: |
  -----BEGIN CERTIFICATE-----
  xxxx  
  -----END CERTIFICATE-----
  -----BEGIN CERTIFICATE-----
  xxxx
  -----END CERTIFICATE-----
```

The collector and query are getting EOF errors / other related errors when trying to connect to the Elasticsearch instance.

### Expected behavior

One should be able to mount a secret with multiple files and then use those files by using either `cmdlineParams` or the environment variables `ES_TLS_XXX`.

One must activate tls.enabled before doing that though, which creates hardcoded mounts and variables.

### Relevant log output

_No response_

### Screenshot

_No response_

### Additional context

I'll try and replicate the error to produce logs again. I've created a fork where the hard-coded values are replaced by variables, so the chart may become more flexible.

### Jaeger backend version

1.45.0

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

Elasticsearch 7.x.x

### Operating system

Linux

### Deployment model

_No response_

### Deployment configs

_No response_

puffitos:
#### What this PR does

Adds the possibility to mount a secret with TLS certificates & keys that contains more than one value. The standard secret mounted always expects the `ca-cert.pem` key to be present, which makes activating TLS complicated in non-default scenarios. 

- Fixes #513 
- Possibly related to #443

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -3,9 +3,9 @@ appVersion: 1.45.0
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
-version: 0.71.17
+version: 0.71.18
 # CronJobs require v1.21
-kubeVersion: '>= 1.21-0'
+kubeVersion: ">= 1.21-0"
 keywords:
   - jaeger
   - opentracing
--- a/charts/jaeger/templates/_helpers.tpl
+++ b/charts/jaeger/templates/_helpers.tpl
@@ -365,7 +365,7 @@ Elasticsearch related environment variables
 - name: ES_TLS_ENABLED
   value: "true"
 - name: ES_TLS_CA
-  value: /es-tls/ca-cert.pem
+  value: {{ .Values.storage.elasticsearch.tls.ca }}
 {{- end }}
 {{- if .Values.storage.elasticsearch.indexPrefix }}
 - name: ES_INDEX_PREFIX
--- a/charts/jaeger/templates/collector-deploy.yaml
+++ b/charts/jaeger/templates/collector-deploy.yaml
@@ -167,8 +167,8 @@ spec:
         {{- end }}
         {{- if .Values.storage.elasticsearch.tls.enabled }}
           - name: {{ .Values.storage.elasticsearch.tls.secretName }}
-            mountPath: "/es-tls/ca-cert.pem"
-            subPath: "ca-cert.pem"
+            mountPath: {{ .Values.storage.elasticsearch.tls.mountPath }}
+            subPath: {{ .Values.storage.elasticsearch.tls.subPath }}
             readOnly: true
         {{- end }}
         {{- if .Values.collector.samplingConfig}}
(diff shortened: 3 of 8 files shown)


puffitos:
Many thanks for the quick reaction 👏

PR #609, opened 2024-10-16
Jaeger v2 with v1

hellspawn679:

* part of #610

#### Checklist

- [ ] Chart Version bumped


charts/jaeger/templates/allinone-deploy.yaml
```diff
@@ -54,10 +54,19 @@ spec:
...
+          image: {{ .Values.v2.image }}
+          {{- end }}
           imagePullPolicy: {{ .Values.allInOne.image.pullPolicy }}
           name: jaeger
           args:
+          {{- if .Values.v2.enabled }}
+            - "--config"
+            - "/etc/jaeger/config.yaml"
+          {{- end -}}
             {{- range $arg := .Values.allInOne.args }}
```
yurishkuro, line 70: I think this should only run if v2 not enabled

charts/jaeger/templates/allinone-agent-svc.yaml
```diff
@@ -1,4 +1,5 @@
 {{- if .Values.allInOne.enabled -}}
+
```
yurishkuro, line 2: if you are not changing the files meaningfully, avoid introducing formatting changes.

yurishkuro (commented):
you need to add a v2 flavor of installation to the CI test

charts/jaeger/templates/allinone-deploy.yaml
hellspawn679: when we enabled v2 we need to pass --config flag as arg and pass the conifg.yaml

charts/jaeger/values.yaml
```diff
@@ -2,6 +2,61 @@
 # This is a YAML-formatted file.
 # Jaeger values are grouped by component. Cassandra values override subchart values
 
+v2:
+  enabled: false
+  image: jaegertracing/jaeger:latest
```
yurishkuro, line 7: 2.0.0-rc2

charts/jaeger/templates/allinone-deploy.yaml
```diff
@@ -54,10 +54,19 @@ spec:
         {{- end }}
           securityContext:
             {{- toYaml .Values.allInOne.securityContext | nindent 12 }}
+          {{- if not .Values.v2.enabled }}
           image: {{ include "allInOne.image" . }}
+          {{- end }}
+          {{- if .Values.v2.enabled }}
```
yurishkuro, line 60: is it possible to do if/else instead of separate ifs?

charts/jaeger/templates/allinone-deploy.yaml
yurishkuro: yes, but we cannot pass any other v1 flags, they will not be recognized

charts/jaeger/templates/allinone-deploy.yaml
```diff
@@ -123,6 +136,11 @@ spec:
         {{- toYaml .Values.allInOne.podSecurityContext | nindent 8 }}
       serviceAccountName: {{ template "jaeger.fullname" . }}
       volumes:
+    {{- if .Values.v2.enabled }}
+        - name: jaeger-v2-config
+          configMap:
+            name: {{ include "jaeger-v2.fullname" . }}
```
yurishkuro, line 142: where is `jaeger-v2.fullname` coming from? does config map name need to be parameterized?

.github/workflows/lint-test-v2.yaml
```diff
@@ -0,0 +1,72 @@
+name: Lint and Test Charts
```
yurishkuro, line 1:
```suggestion
name: jaeger-v2 Lint and Test Charts
```

.github/workflows/lint-test-v2.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+on: pull_request
+
+jobs:
+  lint-test:
```
yurishkuro, line 6:
```suggestion
  lint-test-v2:
```

charts/jaeger/templates/allinone-deploy.yaml
hellspawn679: it is define in _helper.tql in line 734

charts/jaeger/templates/_helpers.tpl
```diff
@@ -719,3 +719,95 @@ Create pull secrets for hotrod image
 {{- define "hotrod.imagePullSecrets" -}}
 {{- include "common.images.renderPullSecrets" (dict "images" (list .Values.hotrod.image) "context" $) -}}
 {{- end }}
+
+
+
+{{- define "jaeger-v2.name" -}}
```
yurishkuro, line 725: Why any of these changes below necessary? Doesn't the chart already have them all for v1? Can we not reuse the same variables?

hellspawn679: yes that is what i planed to do once i was sure that i was going in the right direction with v-2 deployment and config part

yurishkuro:
I see queries going to `Liveness:       http-get http://:14269/` - I don't think this port is available in v2. Health checks should be against :13133

.github/workflows/lint-test-v2.yaml
```diff
@@ -0,0 +1,70 @@
+on: pull_request
```
yurishkuro, line 1:
```suggestion
name: Lint and Test Charts v2

on: pull_request
```

.github/workflows/lint-test-v2.yaml
```diff
@@ -0,0 +1,70 @@
...
+          tar xzf cmctl.tar.gz
+          sudo mv cmctl /usr/local/bin
+          cmctl version
+
+      - name: Check if cert-manager is up
+        run: |
+          cmctl check api --wait=5m
+
+      - name: Run chart-testing (install)
+        run: ct install --debug --config ct.yaml --helm-extra-set-args "--set v2.enabled=true --set provisionDataStore.cassandra=false --set storage.type=memory --set allInOne.enabled=true --set agent.enabled=false --set collector.enabled=false --set query.enabled=false"
```
yurishkuro, line 70:
```suggestion
        run: ct install --debug --config ct.yaml --helm-extra-set-args "--set v2.enabled=true --set provisionDataStore.cassandra=false --set storage.type=memory --set allInOne.enabled=true --set agent.enabled=false --set collector.enabled=false --set query.enabled=false" --helm-extra-args "--debug"
```

charts/jaeger/templates/allinone-deploy.yaml
```diff
@@ -54,10 +54,19 @@ spec:
         {{- end }}
           securityContext:
             {{- toYaml .Values.allInOne.securityContext | nindent 12 }}
+          {{- if not .Values.v2.enabled }}
           image: {{ include "allInOne.image" . }}
+          {{- end }}
+          {{- if .Values.v2.enabled }}
+          image: {{ .Values.v2.image }}
+          {{- end }}
```
yurishkuro, line 62:
```suggestion
          {{- if .Values.v2.enabled }}
          image: {{ .Values.v2.image }}
          {{- else }}
          image: {{ include "allInOne.image" . }}
          {{- end }}
```

charts/jaeger/templates/allinone-deploy.yaml
```diff
@@ -54,10 +54,19 @@ spec:
...
           imagePullPolicy: {{ .Values.allInOne.image.pullPolicy }}
           name: jaeger
           args:
+          {{- if .Values.v2.enabled }}
+            - "--config"
+            - "/etc/jaeger/config.yaml"
+          {{- end -}}
             {{- range $arg := .Values.allInOne.args }}
             - "{{ tpl $arg $ }}"
             {{- end }}
```
yurishkuro, line 72:
```suggestion
          {{- else }}
            {{- range $arg := .Values.allInOne.args }}
            - "{{ tpl $arg $ }}"
            {{- end }}
          {{- end -}}
```

charts/jaeger/templates/allinone-deploy.yaml
yurishkuro: why does the name of config map need to be parameterized?

charts/jaeger/templates/allinone-deploy.yaml
```diff
@@ -123,6 +146,11 @@ spec:
         {{- toYaml .Values.allInOne.podSecurityContext | nindent 8 }}
       serviceAccountName: {{ template "jaeger.fullname" . }}
       volumes:
+    {{- if .Values.v2.enabled }}
+        - name: jaeger-v2-config
+          configMap:
+            name: jaeger-v2-configmap
+    {{- end }}
```
yurishkuro, line 153:
```suggestion
      {{- if .Values.v2.enabled }}
        - name: jaeger-v2-config
          configMap:
            name: jaeger-v2-configmap
      {{- end }}
```

charts/jaeger/values.yaml
```diff
@@ -2,6 +2,62 @@
...
+  # pprof:
+  #   endpoint: 0.0.0.0:1777
+  # zpages:
+  #   endpoint: 0.0.0.0:55679
+
+  jaeger_query:
+    storage:
+      traces: primary_store
+      traces_archive: archive_store
+
```
yurishkuro, line 31:
```suggestion
```

charts/jaeger/values.yaml
```diff
@@ -2,6 +2,62 @@
...
+      archive_store:
+        memory:
+          max_traces: 100000
+
+receivers:
+  otlp:
+    protocols:
+      grpc:
+      http:
+
```
yurishkuro, line 47:
```suggestion
```

charts/jaeger/values.yaml
```diff
@@ -2,6 +2,62 @@
...
+  zipkin:
+
+processors:
+  batch:
+
+exporters:
+  jaeger_storage_exporter:
+    trace_storage: primary_store
+
+
```
yurishkuro, line 60:
```suggestion
# The following settings apply to Jaeger v1 and partially to Jaeger v2

```

.github/workflows/lint-test-v2.yaml
```diff
@@ -0,0 +1,72 @@
+name: Lint and Test Charts v2
```
yurishkuro, line 1: you may want to consider NOT adding a separate workflow, but introducing a matrix in the main workflow with `version: v1 | v2` parameter and only changing the main `ct` step based on that parameter.

Example: https://github.com/jaegertracing/jaeger/blob/acbc0e1f523a36c2080dd4f4544ef6ef66216e86/.github/workflows/ci-e2e-kafka.yml#L24

hellspawn679: yes i did that

charts/jaeger/templates/_helpers.tpl
```diff
@@ -572,6 +572,20 @@ Create pull secrets for all in one image
 {{- include "common.images.renderPullSecrets" (dict "images" (list .Values.allInOne.image) "context" $) -}}
 {{- end }}
 
+{{/*
+Create image name for v2 image
+*/}}
+{{- define "v2.image" -}}
```
pavelnikolov, line 578: again, there shouldn't be any `v2` prefixes. There should be only major chart version bump.

charts/jaeger/templates/allinone-deploy.yaml
```diff
@@ -105,6 +124,10 @@ spec:
         {{- toYaml . | nindent 12 }}
       {{- end }}
           volumeMounts:
+          {{- if .Values.v2.enabled }}
```
pavelnikolov, line 127: I guess we should assume that v2 is always enabled and the only one we support from now on in the main branch. Then there can be v1 branch respectively.

charts/jaeger/values.yaml
```diff
@@ -2,6 +2,66 @@
 # This is a YAML-formatted file.
 # Jaeger values are grouped by component. Cassandra values override subchart values
 
+v2:
```
pavelnikolov, line 5: Let's not try to make this branch support both v1 and v2. We already have a chart which support v1, let's focus 100% on v2 from now on.

.github/workflows/lint-test.yaml
```diff
@@ -69,4 +73,9 @@ jobs:
           cmctl check api --wait=5m
 
       - name: Run chart-testing (install)
-        run: ct install --config ct.yaml
+        run: |
+          if [ "${{ matrix.version }}" == "v1" ]; then
+            ct install --config ct.yaml
+          elif [ "${{ matrix.version }}" == "v2" ]; then
+            ct install --config ct.yaml --helm-extra-set-args "--set v2.enabled=true --set provisionDataStore.cassandra=false --set storage.type=memory --set allInOne.enabled=true --set agent.enabled=false --set collector.enabled=false --set query.enabled=false" 
+          fi
```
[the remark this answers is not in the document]
yurishkuro: @pavelnikolov can you please propose a more concrete plan as far as organizing the repo? How will helm know about different branches?

PR #613, opened 2024-10-26
Jaeger v2 with v1

hellspawn679:

- fixes #

#### Checklist

- [ ] Chart Version bumped


charts/jaeger/values.yaml
```diff
@@ -96,6 +81,75 @@ allInOne:
     fsGroup: 10001
   securityContext: {}
 
+service:
```
yurishkuro, line 84: I don't like how you are mixing the OTEL config sections with chart sections like `provisionDataStore`. Is it possible to wrap the config under parent `config:` element?

charts/jaeger/values.yaml
```diff
@@ -2,30 +2,15 @@
 # This is a YAML-formatted file.
 # Jaeger values are grouped by component. Cassandra values override subchart values
 
-global:
```
yurishkuro, line 5: please minimize the amount of changes in the PR. It looks like you're not really changing this section, but it's being moved to the end so makes it difficult to compare / review.

.github/workflows/lint-test.yaml
```diff
@@ -69,4 +69,5 @@ jobs:
           cmctl check api --wait=5m
 
       - name: Run chart-testing (install)
-        run: ct install --config ct.yaml
+        run: |
+            ct install --config ct.yaml --helm-extra-set-args "--set v2.enabled=true --set provisionDataStore.cassandra=false --set storage.type=memory --set allInOne.enabled=true --set agent.enabled=false --set collector.enabled=false --set query.enabled=false"
```
yurishkuro, line 73: `--set v2.enabled=true` left over?

charts/jaeger/templates/_helpers.tpl
```diff
@@ -719,3 +719,95 @@ Create pull secrets for hotrod image
 {{- define "hotrod.imagePullSecrets" -}}
 {{- include "common.images.renderPullSecrets" (dict "images" (list .Values.hotrod.image) "context" $) -}}
 {{- end }}
+
+
+
+{{- define "jaeger-v2.name" -}}
```
yurishkuro, line 725: we already discussed this - all these settings are already present in the chart without the `v2` name, so please use the existing ones.  I don't think there should be any properties / templates that have `v2` in the name

charts/jaeger/templates/_helpers.tpl
```diff
@@ -719,3 +719,95 @@ Create pull secrets for hotrod image
...
+{{ toYaml .Values.receivers | nindent 6 }}
+{{- end }}
+
+{{- define "jaeger-v2.processorsConfig" -}}
+{{ toYaml .Values.processors | nindent 6 }}
+{{- end }}
+
+{{- define "jaeger-v2.exportersConfig" -}}
+{{ toYaml .Values.exporters | nindent 6 }}
+{{- end }}
```
yurishkuro, line 807: these can be inlined directly in the template, they provide no value

charts/jaeger/templates/allinone-sa.yaml
```diff
@@ -1,14 +0,0 @@
-{{- if .Values.allInOne.enabled -}}
```
yurishkuro, line 1: why is this deleted? all-in-one is still a valid deployment and will need a ServiceAccount

charts/jaeger/templates/jaeger-v2-agent-svc.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.allInOne.enabled -}}
```
yurishkuro, line 1: why do we need to rename all these files? Please keep the same names - there should be no files with `v2` in the name.

.github/workflows/lint-test.yaml
hellspawn679: oh i changes v2 to jaeger will fix it now

charts/jaeger/templates/allInOne-config.yaml
```diff
@@ -0,0 +1,27 @@
+{{- if .Values.allInOne.enabled -}}
```
yurishkuro, line 1: We can run jaeger-v2 binary without config and it will also work as all-in-one. Do we need this file? If people want to customize then can provide a custom config.

hellspawn679: yes we need it because if the user provides there own config file it will update this config template and then pass it to jaeger-v2 deployment. then how are we gonna decide that a user wanted to give there own config if we don't pass it or need to create a enabled flag which then again increase the things a user need to learn so we can simply pass it like this

yurishkuro: Don't follow your explanation. We must support the ability for the user to provide their custom config that completely replaces the default one. Given that as a requirements, why do we need a built-in "all in one" config in the chart if we already have it bundled with the binary directly? What use cases is that config meant to solve?

charts/jaeger/templates/allInOne-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: jaeger-configmap
+  namespace: {{ include "jaeger.namespace" . }}
+  labels:
+        {{- include "jaeger.labels" . | nindent 4 }}
```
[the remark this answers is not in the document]
hellspawn679: could you explain what do yo mean by "normal indentation" i used this format because that is how labels are indented thought-out the chart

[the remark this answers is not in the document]
yurishkuro: I showed example

charts/jaeger/templates/allInOne-config.yaml
hellspawn679: well currently i am using this config map using 
```
  args:
             - "--config"
             - "/etc/jaeger/config.yaml"
```
if i decide not to use it i would have to add a extra if condition to this arg 
```
  args:
         {{- if values.config.enabled }}
             - "--config"
             - "/etc/jaeger/config.yaml"
         {{- end}}
```
this will unnecessary increase the flag plus we can use this config later for es and Cassandra  too.

yurishkuro: Once again, is it a REQUIREMENT that user can provide their own config file. How do you plan to support that?

hellspawn679: user can simply pass pass there config in eg . if he/she have a `test-config.yaml` file 
```
config:
  service:
    extensions: [jaeger_storage, jaeger_query, healthcheckv2]
    pipelines:
      traces:
        receivers: [otlp, jaeger, zipkin]
        processors: [batch]
        exporters: [jaeger_storage_exporter]

  extensions:
    healthcheckv2:
      use_v2: true
      http:
        endpoint: 0.0.0.0:13133
    jaeger_query:
      storage:
        traces: primary_store
        traces_archive: archive_store

    jaeger_storage:
      backends:
        primary_store:
          memory:
            max_traces: 90000
        archive_store:
          memory:
            max_traces: 90000

  receivers:
    otlp:
      protocols:
        grpc:
        http:
    jaeger:
      protocols:
        grpc:
    zipkin:

  processors:
    batch:

  exporters:
    jaeger_storage_exporter:
      trace_storage: primary_store
```
the user can pass this config file 
```
helm template charts/jaeger     --set provisionDataStore.cassandra=false     --set storage.type=memory     --set allInOne.enabled=true     --set agent.enabled=false     --set collector.enabled=false     --set query.enabled=false  -f config-test.yaml 
```

the output will look like this it will override the default values 
```
        extensions:
      
            healthcheckv2:
              http:
                endpoint: 0.0.0.0:13133
              use_v2: true
            jaeger_query:
              storage:
                traces: primary_store
                traces_archive: archive_store
            jaeger_storage:
              backends:
                archive_store:
                  memory:
                    max_traces: 90000
                primary_store:
                  memory:
                    max_traces: 90000
```

yurishkuro: what does `-f config-test.yaml ` mean? is this a helm flag? would it not override values.yaml completely, not just collector config?

yurishkuro: > if i decide not to use it i would have to add a extra if condition to this arg

no, you can use `allinone.enabled` as condition

hellspawn679: if allinOne is enabled then i deployment will always run and along with it the arguments will always be passed which means we will need a config map to pass with the arg

charts/jaeger/templates/allInOne-config.yaml
```diff
@@ -0,0 +1,12 @@
+{{ if .Values.userconfig }}
```
yurishkuro, line 1: I think this file should be called `user-config.yaml`, it's not specific to all-in-one.

charts/jaeger/values.yaml
```diff
@@ -2,6 +2,61 @@
 # This is a YAML-formatted file.
 # Jaeger values are grouped by component. Cassandra values override subchart values
 
+# Will use for user provided config
+userconfig:
+
+config:
```
yurishkuro, line 8: can we now remove this section?

hellspawn679: I am keep it for es i will remove it if I don't need it later

charts/jaeger/templates/allInOne-config.yaml
hellspawn679: i used that only in the start but I got this error 
`Error: INSTALLATION FAILED: parse error at (jaeger/templates/allInOne-config.yaml:1): bad character U+002D '-'`
the problem is helm chart don't allow '-' there  is workaround we can use 
`{{index .Values.user-config }}`
but with if or any other condition statements it throw this error 
`Error: INSTALLATION FAILED: parse error at (jaeger/templates/allInOne-config.yaml:1): unexpected <if> in operand`

yurishkuro: I made this change to the template
```diff
diff --git a/charts/jaeger/templates/allInOne-config.yaml b/charts/jaeger/templates/allInOne-config.yaml
index eca21f2..e473618 100644
--- a/charts/jaeger/templates/allInOne-config.yaml
+++ b/charts/jaeger/templates/allInOne-config.yaml
@@ -8,5 +8,5 @@ metadata:
     {{- include "jaeger.labels" . | nindent 4 }}
 data:
   config.yaml: |
-        {{ .Values.userconfig | nindent 8 }}
+        {{- .Values.userconfig | nindent 4 }}
 {{- end }}
```

created mock config file
```
$ cat myconfig.yaml
hello:
  world: true
```
and ran a test 
```
$ helm template --set-file userconfig=myconfig.yaml charts/jaeger > test-output.txt
```

which resulted in reasonably looking resource file:
```yaml
# Source: jaeger/templates/allInOne-config.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: jaeger-configmap
  namespace: default
  labels:
    helm.sh/chart: jaeger-4.0.0
    app.kubernetes.io/name: jaeger
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/version: "2.0.0-rc2"
    app.kubernetes.io/managed-by: Helm
data:
  config.yaml: |
    hello:
      world: true
```

charts/jaeger/templates/allinone-deploy.yaml
```diff
@@ -123,6 +133,11 @@ spec:
         {{- toYaml .Values.allInOne.podSecurityContext | nindent 8 }}
       serviceAccountName: {{ template "jaeger.fullname" . }}
       volumes:
+      {{ if .Values.userconfig }}
+         - name: jaeger-config
+           configMap:
+            name: jaeger-configmap
```
yurishkuro, line 139: please clean-up all the naming. Not jaeger-config, not jaeger-configmap, not `allInOne-config.yaml`, not `config.yaml` - everything should be using a single name `user-config`

hellspawn679: done should i convert it to PR

yurishkuro:
I merged changes from main into v2, let's resolve conflicts

.github/workflows/lint-test.yaml
```diff
@@ -58,32 +47,7 @@ jobs:
...
-            --set query.enabled=false"
-  test-with-elasticsearch:
-    runs-on: ubuntu-latest
-    steps:
-        - uses: actions/checkout@v2
-          with:
-            fetch-depth: 0
-
-        - uses: ./.github/actions/prepare-k8s
+            ct install --config ct.yaml --helm-extra-set-args "--set provisionDataStore.cassandra=false --set storage.type=memory --set allInOne.enabled=true --set agent.enabled=false --set collector.enabled=false --set query.enabled=false"
```
yurishkuro, line 52: this line is hard to read, please use the same formatting as in the deleted sections, one flag per line

yurishkuro (commented):
please delete empty es.txt

.github/workflows/lint-test.yaml
```diff
@@ -38,17 +38,6 @@ jobs:
...
-    steps:
-      - uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
-
-      - uses: ./.github/actions/prepare-k8s
-
-      - name: Run cassandra-chart-testing (install)
-        run: ct install --config ct.yaml
   test-with-allInOne:
```
yurishkuro, line 41:
```suggestion
  test-with-all-in-one:
```

yurishkuro (commented):
couple small things, ready to merge otherwise

charts/jaeger/templates/user-config.yaml
```diff
@@ -0,0 +1,12 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: user-config
+  namespace: {{ include "jaeger.namespace" . }}
+  labels:
+    {{- include "jaeger.labels" . | nindent 4 }}
+data:
+  user-config.yaml: |
+        {{- .Values.userconfig | nindent 4 }}
```
yurishkuro, line 11:
```suggestion
  user-config.yaml: |
    {{- .Values.userconfig | nindent 4 }}
```

yurishkuro (approved):

PR #614, opened 2024-10-26
Fix hostname for Elasticsearch and add CI test

hellspawn679:

- fixes #

#### Checklist

- [ ] Chart Version bumped


charts/jaeger/templates/_helpers.tpl
```diff
@@ -268,7 +268,11 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
 */}}
 {{- define "elasticsearch.client.url" -}}
 {{- $port := .Values.storage.elasticsearch.port | toString -}}
-{{- printf "%s://%s:%s" .Values.storage.elasticsearch.scheme .Values.storage.elasticsearch.host $port }}
+{{- $host := .Values.storage.elasticsearch.host }}
+{{- if .Values.provisionDataStore.elasticsearch }}
+{{- $host = printf "%s-elasticsearch" .Release.Name }}
+{{- end }}
```
yurishkuro, line 274: this looks like it overrides the user-provided host. Shouldn't it be more like

```suggestion
{{- $host := printf "%s-elasticsearch" .Release.Name }}
{{- if .Values.storage.elasticsearch.host}}
  {{- $host = .Values.storage.elasticsearch.host }}
{{- end }}
```

charts/jaeger/templates/collector-deploy.yaml
```diff
@@ -32,6 +32,12 @@ spec:
         {{- toYaml .Values.collector.podLabels | nindent 8 }}
 {{- end }}
     spec:
+      {{- if .Values.provisionDataStore.elasticsearch }}
+      initContainers:
+        - name: elasticsearch-checker
+          image: curlimages/curl
+          command: ["sh", "-c", "until curl -s http://{{ .Release.Name }}-elasticsearch:9200/_cluster/health?wait_for_status=green; do echo waiting for Elasticsearch; sleep 5; done;"]
```
yurishkuro, line 39: I think you want to use `elasticsearch.client.url` from _helpers, not to compose the host name again

charts/jaeger/templates/_helpers.tpl
hellspawn679: there is always a default value present in `values.yaml `  so it will always override the `.Values.provisionDataStore.elasticsearch`
and additionally it doesn't matter what you pass to `.Values.storage.elasticsearch.host` since in provision storage it is gonna use it's own chart so changing host name will not affect the name of provision es service

yurishkuro: it doesn't make sense to have a `host` property in values.yaml if it has no impact. Should we set it to null then? Because in case the user manages their own Elastic cluster they may still need to provide a different name.

charts/jaeger/templates/collector-deploy.yaml
```diff
@@ -32,6 +32,20 @@ spec:
...
+      {{- if .Values.provisionDataStore.elasticsearch }}
+      initContainers:
+        - name: elasticsearch-checker
+          image: curlimages/curl
+          command: 
+          - sh
+          - "-c"
+          - |
+            url="{{ include "elasticsearch.client.url" . }}"
+            until curl -s "$url/_cluster/health?wait_for_status=green"; do
```
yurishkuro, line 44: `until` checks process exit code. Should you not be checking HTTP status code instead?

yurishkuro (approved):

.github/workflows/es.yaml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Set up cmctl
+        run: |
+          curl -sSL -o cmctl.tar.gz https://github.com/cert-manager/cert-manager/releases/download/v1.6.1/cmctl-linux-amd64.tar.gz
+          tar xzf cmctl.tar.gz
+          sudo mv cmctl /usr/local/bin
+          cmctl version
+      - name: Check if cert-manager is up
+        run: |
+          cmctl check api --wait=5m
+      - name: Run chart-testing (install)
```
yurishkuro, line 67: Suggestion: we can move all steps up to this point into a shared helper action ([example](https://github.com/jaegertracing/jaeger/tree/main/.github/actions/setup-branch)) and not repeat them in each workflow. Then instead of defining a new workflow file you can add an extra job in the main workflow

.github/workflows/lint-test.yaml
```diff
@@ -68,5 +68,27 @@ jobs:
         run: |
           cmctl check api --wait=5m
 
-      - name: Run chart-testing (install)
+      - name: Run cassandra-chart-testing (install)
         run: ct install --config ct.yaml
+
+      - name: Run allInOne-chart-testing (install)
```
yurishkuro, line 74: please rebase and split these into separate jobs (see #615)

yurishkuro (approved):
LGTM

PR #617, opened 2024-11-08
[jaeger-v2] Add support for Elasticsearch

hellspawn679:
#### What this PR does
added support for elasticsearch in jaeger-v2 helm chart

- fixes add support for elasticsearch for jaeger v2

#### Checklist

- [x] Chart Version bumped


yurishkuro:
please write complete PR title and description

charts/jaeger/Chart.yaml
```diff
@@ -4,6 +4,8 @@ description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
 version: 4.0.0
+annotations:
+  Jaegerv1Version: "1.53.0"
```
yurishkuro, line 8:
```suggestion
  Jaegerv1Version: "1.62.0"
```
use latest where agent was present

charts/jaeger/templates/collector-svc.yaml
```diff
@@ -61,7 +61,7 @@ spec:
     targetPort: {{ .Values.collector.service.otlp.http.name }}
 {{- end }}
   - name: {{ .Values.collector.service.admin.name }}
```
yurishkuro, line 63: What was the meaning of "admin"? In v2 there is no single admin port.

charts/jaeger/templates/es-user-config.yaml
```diff
@@ -0,0 +1,62 @@
...
+        storage:
+          traces: some_storage
+          traces_archive: another_storage
+
+
+      jaeger_storage:
+        backends:
+          some_storage:
+            elasticsearch:
+              index_prefix: {{- .Values.storage.elasticsearch.index_prefix | quote | indent 2 }}
```
yurishkuro, line 35: I don't like this. `Values.storage.elasticsearch.index_prefix` is a separate domain-specific configuration language that exists only in the Operator - why do we need it? The users can provide their own custom config if they want to customize these things. If we still want to allow overriding specific fields of the built-in configuration, then those overrides should be using the OTEL config syntax, e.g. `Values.config.extensions.jaeger_storage...etc...`

user-config.yaml
```diff
@@ -0,0 +1,45 @@
+service:
```
yurishkuro, line 1: should this file be called `user-config.yaml`? It's ES-specific.

charts/jaeger/templates/es-user-config.yaml
```diff
@@ -0,0 +1,62 @@
+{{ if (and (not .Values.userconfig) .Values.provisionDataStore.elasticsearch) }}
```
yurishkuro, line 1: should this file be called "user config"? It looks like a built-in config. Also it's not even a config, but a configmap, so let's call it that.

user-config.yaml
hellspawn679: No, if the user provides their own config, this will be used instead of `es-user-config.yaml`.

charts/jaeger/templates/es-user-config.yaml
hellspawn679: I wanted to keep the flag syntax same as jaeger-v1 so if the users wants to run it with old flag he can set it that way and helm will use `es-user-conifg.yaml `for that but if the user wants to provide there own config he can use the `-set-file` flag in that case we will use `user-config.yaml`

charts/jaeger/templates/es-user-config.yaml
hellspawn679: Ok, I will rename it to `es-configmap.yaml `

charts/jaeger/templates/collector-svc.yaml
hellspawn679: So what should I name it in `values.yaml`

.github/workflows/lint-test.yaml
```diff
@@ -50,11 +50,32 @@ jobs:
       - name: Run chart-testing (install)
         run: |
           ct install --config ct.yaml \
-            --helm-extra-set-args "
+              --helm-extra-set-args "
```
pavelnikolov, line 53: Are the quites in the back intentional? Also should you include end of line, i.e. `\`?

.github/workflows/lint-test.yaml
```diff
@@ -50,11 +50,32 @@ jobs:
...
+        - uses: actions/checkout@v2
+          with:
+            fetch-depth: 0
+
+        - uses: ./.github/actions/prepare-k8s
+
+        - name: Run chart-testing (install)
+          run: |
+            ct install --config ct.yaml \
+                --helm-extra-set-args "
```
pavelnikolov, line 72: Again, end of line `\` missing.

charts/jaeger/templates/collector-svc.yaml
yurishkuro: This specific port is healthcheck

charts/jaeger/templates/es-configmap.yaml
```diff
@@ -0,0 +1,62 @@
...
+        storage:
+          traces: some_storage
+          traces_archive: another_storage
+
+
+      jaeger_storage:
+        backends:
+          some_storage:
+            elasticsearch:
+              index_prefix: {{- .Values.config.extensions.jaeger_storage.backends.elasticsearch.index_prefix | quote | indent 2 }}
```
yurishkuro, line 35: why do you need to use `.Values.config.extensions.jaeger_storage.backends.elasticsearch. index_prefix ` reference here?

charts/jaeger/values.yaml
```diff
@@ -32,6 +32,12 @@ config:
 
     jaeger_storage:
       backends:
+        elasticsearch:
```
yurishkuro, line 35: this is not a valid syntax. It should be 
```
backends:
  {some name}:
    elasticsearch: // this refers to the type of storage
```

user-config.yaml
yurishkuro: is this file meant to be an example config?

yurishkuro (commented):
please document in the README how the user is expected to use the chart with ES, using different options such as providing params via `--set` or a full config via `--set-file`

.github/workflows/lint-test.yaml
```diff
@@ -50,11 +50,32 @@ jobs:
...
+          with:
+            fetch-depth: 0
+
+        - uses: ./.github/actions/prepare-k8s
+
+        - name: Run chart-testing (install)
+          run: |
+            ct install --config ct.yaml \
+                --helm-extra-set-args " \
+                --set provisionDataStore.cassandra=false \
```
yurishkuro, line 73: this should be default, can we avoid setting it?

.github/workflows/lint-test.yaml
```diff
@@ -50,11 +50,32 @@ jobs:
       - name: Run chart-testing (install)
         run: |
           ct install --config ct.yaml \
-            --helm-extra-set-args "
+              --helm-extra-set-args "  \
               --set provisionDataStore.cassandra=false \
```
yurishkuro, line 54: is this flag relevant to memory storage?

charts/jaeger/README-v2.md
```diff
@@ -0,0 +1,109 @@
...
+
+### 1. Jaeger All-in-One Mode
+
+Jaeger’s all-in-one mode combines the Jaeger Agent, Collector, and Query into a single pod for simplicity. Use this mode for testing or lightweight deployments.
+
+#### **Command to Run All-in-One**
+
+```bash
+helm install <chart_name> charts/jaeger \
+    --set provisionDataStore.cassandra=false \
```
yurishkuro, line 19: is it releavant?

.github/workflows/lint-test.yaml
```diff
@@ -50,11 +50,32 @@ jobs:
...
               --set storage.type=memory \
               --set allInOne.enabled=true \
               --set agent.enabled=false \
               --set collector.enabled=false \
               --set query.enabled=false "
```
yurishkuro, line 59: I think these should automatically be treated as false if `allInOne` is true

charts/jaeger/templates/es-configmap.yaml
```diff
@@ -0,0 +1,62 @@
...
+      jaeger_query:
+        storage:
+          traces: primary_store
+          traces_archive: archive_store
+
+
+      jaeger_storage:
+        backends:
+          primary_store:
+            elasticsearch:
```
yurishkuro, line 34: I don't mind merging like this as a milestone, but I think this file should not be specific to ES only, it can use the flags indicating which storage type the user selected and creating the corresponding configs for that storage. Otherwise we would have to duplicate this overall config for each storage type.

charts/jaeger/templates/es-configmap.yaml
```diff
@@ -0,0 +1,62 @@
...
+              username: {{- .Values.config.extensions.jaeger_storage.backends.archive_store.elasticsearch.user | quote | indent 2 }}
+              {{- if .Values.config.extensions.jaeger_storage.backends.archive_store.elasticsearch.password }}
+              password: {{- .Values.config.extensions.jaeger_storage.backends.archive_store.elasticsearch.password | quote | indent 2 }}
+              {{- end }}
+
+    receivers:
+      otlp:
+        protocols:
+          grpc:
+          http:
```
yurishkuro, line 54: we need to set endpoints to 0.0.0.0:portNum, see https://github.com/jaegertracing/jaeger/issues/6209

charts/jaeger/values.yaml
```diff
@@ -33,9 +33,19 @@ config:
     jaeger_storage:
```
yurishkuro, line 33: help me understand - this `config` structure is not actually used as a config, right? It only provides the shape for the properties that looks like OTEL config?

charts/jaeger/values.yaml
```diff
@@ -566,7 +576,7 @@ collector:
         # name: otlp-http
         # port: 4318
         # nodePort:
-    admin:
+    healthCheck:
       name: admin
```
yurishkuro, line 580: do we need to change the name away from `admin`? It's an inconsistency

user-config.yaml
yurishkuro: ?

charts/jaeger/templates/user-config.yaml
```diff
@@ -1,3 +1,4 @@
+# this is a user-provided config file that will be mounted into the all-in-one deployments
```
yurishkuro, line 1:
```suggestion
# Generates a config map from a file provided by user via `--set-file userconfig=`
```

charts/jaeger/templates/query-svc.yaml
```diff
@@ -24,7 +24,7 @@ spec:
     protocol: TCP
     targetPort: grpc
   - name: {{ .Values.query.service.admin.name }}
```
yurishkuro, line 26:
```suggestion
  - name: {{ .Values.query.service.healthcheck.name }}
```

charts/jaeger/templates/agent-ds.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.agent.enabled -}}
+{{- if and .Values.agent.enabled (not .Values.allInOne.enabled)  -}}
```
yurishkuro, line 1: > `.Values.allInOne.enabled`

I don't think you should have made this change. I simply meant to change the defaults in values.yaml, not to introduce these massive changes across the chart.

charts/jaeger/values.yaml
```diff
@@ -33,17 +33,27 @@ config:
...
           memory:
             max_traces: 100000
 
   receivers:
     otlp:
       protocols:
-        grpc:
-        http:
+        grpc: 0.0.0.0:4317
+        http: 0.0.0.0:4318
```
yurishkuro, line 56:
```suggestion
        grpc:
          endpoint: 0.0.0.0:4317
        http:
          endpoint: 0.0.0.0:4318
```

charts/jaeger/templates/collector-deploy.yaml
```diff
@@ -123,7 +126,7 @@ spec:
         - containerPort: {{ .Values.collector.service.http.port }}
           name: http
           protocol: TCP
-        - containerPort: 14269
+        - containerPort: 13133
           name: admin
```
yurishkuro, line 130: should this not be called `healthcheck`? Is this name used anywhere?

yurishkuro (approved):

PR #621, opened 2024-12-01
[jaeger] Make curl image respect global.imageRegistry

FieteO:
#### What this PR does
This PR makes the curl image that is used for elasticsearch respect `global.imageRegistry`, such that:
```sh
helm template --set provisionDataStore.elasticsearch=true . | grep curlimages
          image: curlimages/curl
          image: curlimages/curl
```
```sh
helm template --set provisionDataStore.elasticsearch=true,global.imageRegistry=example.com . | grep curlimages
          image: example.com/curlimages/curl
          image: example.com/curlimages/curl
```

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/templates/_helpers.tpl
+++ b/charts/jaeger/templates/_helpers.tpl
@@ -718,6 +718,18 @@ Create image name for hotrod image
 {{- end -}}
 
 {{/*
+Define curl image declaration
+*/}}
+{{- define "curl.image" -}}
+{{- $image := "curlimages/curl" -}}
+{{- if .Values.global.imageRegistry -}}
+{{ .Values.global.imageRegistry }}/{{ $image }}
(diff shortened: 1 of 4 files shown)


charts/jaeger/templates/_helpers.tpl
```diff
@@ -717,6 +717,18 @@ Create image name for hotrod image
 {{- include "renderImage" ( dict "imageRoot" .Values.hotrod.image "context" $ ) -}}
 {{- end -}}
 
+{{/*
+Define curl image declaration
+*/}}
+{{- define "curl.image" -}}
+{{- $image := "curlimages/curl" -}}
```
pavelnikolov, line 724: Should users be able to override the image as well?

pavelnikolov (approved):

PR #671, opened 2025-12-09
 upgrade to Jaeger v2, remove legacy components, and fix storage config

danish9039:
**Title:**  upgrade to Jaeger v2, support OS/ES/Cassandra, and remove legacy components

**Description:**
- This PR upgrades the Jaeger Helm chart to support **Jaeger v2 (v2.13.0)** exclusively.
- It removes all legacy v1 microservices (Agent, Collector, Query, Ingester) and refactors the configuration generation to support multiple storage backends natively.

**Key Changes:**

*   **Upgrade to Jaeger v2:**
    *   Bumps Chart version to `4.0.1` and App version to `2.13.0`.
    *   Updates [allinone-deploy.yaml](cci:7://file:///home/danish/opensource/helm-charts/charts/jaeger/templates/allinone-deploy.yaml:0:0-0:0) to use the Jaeger v2 binary and mount a dynamically generated `user-config.yaml`.
*   **Storage Backend Support :**
    *   **Elasticsearch & OpenSearch:** Refactored [es-configmap.yaml](cci:7://file:///home/danish/opensource/helm-charts/charts/jaeger/templates/es-configmap.yaml:0:0-0:0) to generate correct Jaeger v2 / OpenTelemetry `jaeger_storage` configuration. Supports separate `primary_store` and `archive_store`, TLS configuration (`insecure` support), and authentication via `auth.basic` block. Verified compatibility with both Elasticsearch and OpenSearch clusters.
    *   **Cassandra:** Updated [es-configmap.yaml](cci:7://file:///home/danish/opensource/helm-charts/charts/jaeger/templates/es-configmap.yaml:0:0-0:0) and [cassandra-schema-job.yaml](cci:7://file:///home/danish/opensource/helm-charts/charts/jaeger/templates/cassandra-schema-job.yaml:0:0-0:0) to decouple from legacy flags.
    *   **Memory:** Default in-memory storage remains supported for development.
*   **Legacy Cleanup:**
    *   Removed all templates, values, and helpers related to Jaeger v1 microservices (`agent`, `collector`, `query`, `ingester`).
    *   Removed [README-v2.md](cci:7://file:///home/danish/opensource/helm-charts/charts/jaeger/README-v2.md:0:0-0:0) and cleaned up [README.md](cci:7://file:///home/danish/opensource/helm-charts/charts/jaeger/README.md:0:0-0:0) to remove obsolete flags and sections.
*   **Service Routing:**
    *   Updated `allinone-agent-svc`, `allinone-collector-svc`, and `allinone-query-svc` to route all traffic (OTLP, Zipkin, Jaeger, UI) to the single Jaeger v2 All-in-One pod.

**Verification:**
runtime testing was performed on a **Kind cluster** across three scenarios:

1.  **In-Memory:**
    *   `helm install ...`
    *   Result: Pod started successfully in default configuration.
2.  **Cassandra:**
    *   `helm install ... --set provisionDataStore.cassandra=true --set storage.type=cassandra`
    *   Result: Schema job completed, Jaeger authenticated and connected successfully.
3.  **OpenSearch (via Elasticsearch compatibility):**
    *   `helm install ... --set storage.type=elasticsearch ...` (pointing to a manually deployed OpenSearch cluster with TLS and Auth).
    *   Result: Jaeger authenticated using `admin`/`StrongPassword123!`, accepted self-signed TLS certificates, and successfully initialized the storage backend.

**Breaking Changes:**
*   This chart no longer supports Jaeger v1.
*   `collector.enabled`, `query.enabled`, `agent.enabled` flags are removed.
*   Users must migrate to the v2 All-in-One architecture .

**Checklist:**
- [x] Chart Version bumped
- [x] Legacy components removed
- [x] Storage backends (Memory, Cassandra, ES/OpenSearch) verified
- [x] Documentation updated

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-appVersion: "2.13.0"
+appVersion: "2.12.0"
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
--- a/charts/jaeger/README.md
+++ b/charts/jaeger/README.md
@@ -41,6 +41,23 @@ By default, the chart deploys:
 - Jaeger All-in-One Deployment (combines collector, query, and internal components)
 - In-memory storage (non-persistent)
 
+### Overriding the Jaeger Version
+
+The chart version maps to a specific Jaeger version (appVersion). You can override this to use a different Jaeger image version without upgrading the chart:
+
+```bash
+helm install jaeger jaegertracing/jaeger \
+  --set allInOne.image.tag=2.13.0
+```
+
+Or via a values file:
+
+```yaml
+allInOne:
+  image:
+    tag: "2.13.0"
+```
+
 ## Configuration
 
 Jaeger v2 uses a YAML-based configuration format built on the OpenTelemetry Collector framework. The configuration is defined in the `config` section of the values file.
--- a/charts/jaeger/templates/NOTES.txt
+++ b/charts/jaeger/templates/NOTES.txt
@@ -2,12 +2,9 @@
 ###################################################################
 ### IMPORTANT: Ensure that storage is explicitly configured     ###
 ### Default storage options are subject to change.              ###
-###                                                             ###
-### IMPORTANT: The use of <component>.env: {...} is deprecated. ###
-### Please use <component>.extraEnv: [] instead.                ###
 ###################################################################
 
-🚀 Congratulations on successfully installing / upgrading Jaeger v{version}!
+🚀 Congratulations on successfully installing / upgrading Jaeger v{{ .Chart.AppVersion }}!
 
 📊 You can log into the Jaeger Query UI here:
 
--- a/charts/jaeger/templates/allinone-agent-svc.yaml
+++ b/charts/jaeger/templates/allinone-agent-svc.yaml
@@ -1,7 +1,7 @@
 # This service provides backward compatibility for clients expecting
 # the legacy Jaeger Agent UDP endpoints (5775, 6831, 6832).
 # It routes traffic to the All-in-One pod.
-{{- if .Values.allInOne.enabled -}}
+{{ if .Values.allInOne.enabled -}}
 apiVersion: v1
 kind: Service
 metadata:
--- a/charts/jaeger/templates/storage-configmap.yaml
+++ b/charts/jaeger/templates/storage-configmap.yaml
@@ -45,7 +45,7 @@ data:
               {{- end }}
               tls:
                 insecure: {{ .Values.storage.elasticsearch.tls.insecure | default false }}
-                {{- if .Values.storage.elasticsearch.tls.ca }}
+                {{- if and .Values.storage.elasticsearch.tls.enabled .Values.storage.elasticsearch.tls.ca }}
                 ca_file: {{ .Values.storage.elasticsearch.tls.ca }}
                 {{- end }}
             {{- else if or .Values.provisionDataStore.cassandra (eq .Values.storage.type "cassandra") }}
@@ -82,7 +82,7 @@ data:
               {{- end }}
               tls:
                 insecure: {{ .Values.storage.elasticsearch.tls.insecure | default false }}
-                {{- if .Values.storage.elasticsearch.tls.ca }}
+                {{- if and .Values.storage.elasticsearch.tls.enabled .Values.storage.elasticsearch.tls.ca }}
                 ca_file: {{ .Values.storage.elasticsearch.tls.ca }}
                 {{- end }}
             {{- else }}


charts/jaeger/templates/allinone-agent-svc.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ template "jaeger.agent.name" . }}
+  name: {{ include "jaeger.fullname" . }}-agent
```
yurishkuro, line 8: what is the role of this file?

allinone-agent-svc.yaml

If it's still needed please add a header comment

charts/jaeger/templates/allinone-collector-svc.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ template "jaeger.collector.name" . }}
+  name: {{ include "jaeger.fullname" . }}-collector
```
yurishkuro, line 5: same question as above

charts/jaeger/templates/allinone-deploy.yaml
```diff
@@ -62,7 +62,7 @@ spec:
           imagePullPolicy: {{ .Values.allInOne.image.pullPolicy }}
           name: jaeger
           args:
-            {{- if .Values.userconfig }}
+            {{- if or .Values.userconfig .Values.provisionDataStore.elasticsearch .Values.provisionDataStore.cassandra }}
```
yurishkuro: why is creating an external config dependent on storage configuration? Specifically, is it possible to run with ES or Cassandra without user-provided config file?

charts/jaeger/templates/allinone-agent-svc.yaml
yurishkuro: it looks like it creates a "service" for UDP ports, but why do we need a dedicated name for it like "xx-agent" and "xx-collector" below? These terms are now mostly meaningless in Jaeger v2 architecture, every port that receives the data is a "collector" function.

charts/jaeger/templates/cassandra-schema-job.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.collector.enabled -}}
```
yurishkuro, line 1: do we still need this job? In the `Schema` section of Cassandra config we have the flag to create the schema automatically from the main Jaeger binary
```
	// CreateSchema tells if the schema ahould be created during session initialization based on the configs provided
	CreateSchema bool `mapstructure:"create" valid:"optional"`
```

charts/jaeger/templates/storage-configmap.yaml
```diff
@@ -27,33 +27,85 @@ data:
...
+                  password: {{ .Values.config.extensions.jaeger_storage.backends.primary_store.elasticsearch.password | quote }}
+                  {{- end }}
               {{- end }}
+              tls:
+                insecure: {{ .Values.storage.elasticsearch.tls.insecure | default false }}
+                {{- if .Values.storage.elasticsearch.tls.ca }}
+                ca_file: {{ .Values.storage.elasticsearch.tls.ca }}
+                {{- end }}
+            {{- else if .Values.provisionDataStore.cassandra }}
+            cassandra:
```
yurishkuro, line 52: the file is called es-configmap, should it be called storage-configmap instead?

charts/jaeger/templates/NOTES.txt
yurishkuro, line 5: what is the purpose of this NOTES.txt file?

charts/jaeger/README.md
yurishkuro: Could we add a section at the beginning of README that talks about the different deployment shapes?
  * all-in-one is always a single instance, correct?
  * can all-in-one still be used with a persistent storage?
  * if I don't want all-in-one, what does it look like? A typical expectation would be a group of `collector` instances, and `query` instances.
  * is collector-kafka-ingester setup supported?

It is ok if some of these are not supported, we can explicitly say so, I just want the README to correctly reflect the current state.

yurishkuro:
```suggestion
See [Customizing the Chart Before Installing](https://helm.sh/docs/intro/using_helm/#customizing-the-chart-before-installing). To see all configurable options with detailed comments, visit the chart's [values.yaml](https://github.com/jaegertracing/helm-charts/blob/main/charts/jaeger/values.yaml), or run these configuration commands:
```

yurishkuro, line 130:
```suggestion

See more configuration examples in https://github.com/jaegertracing/jaeger/tree/main/cmd/jaeger
```

charts/jaeger/templates/cassandra-schema-job.yaml
danish9039: That is the long-term goal, yes. However, during verification, I discovered a bug in the current Jaeger v2 binary where `schema.create: true` generates invalid `CQL `when a `datacenter `is configured (specifically for `NetworkTopologyStrategy`).  

https://github.com/jaegertracing/jaeger/blob/ca5482e217c20063fc6f96948c2aca646466c6c6/internal/storage/v1/cassandra/schema/schema.go#L50

I have verified this failure locally and raised a PR for this . https://github.com/jaegertracing/jaeger/pull/7726

Until that fix is merged and released in a new Jaeger binary version, this Schema Job is strictly required to ensure schemas are created correctly . We can deprecate this job in a future release once the binary fix is available."

charts/jaeger/templates/storage-configmap.yaml
danish9039: > the file is called es-configmap, should it be called storage-configmap instead?

renamed `es-configmap` to `storage-configmap`

charts/jaeger/README.md
danish9039: done , please check Readme

charts/jaeger/templates/allinone-deploy.yaml
danish9039: Yes, valid usage without a user-provided config file is fully supported

charts/jaeger/templates/NOTES.txt
danish9039: this file is a standard Helm template that provides instructions to the user immediately after a successful helm install or upgrade in terminal

charts/jaeger/templates/allinone-agent-svc.yaml
danish9039: i have not deleted these services because it will likely cause a braking change for existing users , if we are ok with that we can delete these services

charts/jaeger/Chart.yaml
```diff
@@ -1,11 +1,9 @@
 apiVersion: v2
-appVersion: 2.0.0-rc2
+appVersion: "2.13.0"
```
yurishkuro: if possible (i.e. no breaking config changes) then I would like to start with v2.12 so that after we merge this we can also test the auto-upgrade workflow

yurishkuro: another question - can a user override this version without having to upgrade the chart itself? I understand it may not work if there are breaking config changes, but it's still nice to give that flexibility _and document it_. They can always work around breaking config changes by supplying the full custom config.

charts/jaeger/templates/NOTES.txt
yurishkuro: so is this relevant?

> `The use of <component>.env: {...} is deprecated`

charts/jaeger/templates/NOTES.txt
```diff
@@ -9,22 +9,12 @@
 
 You can log into the Jaeger Query UI here:
```
yurishkuro: Can we add something like 

```suggestion
🚀 Congratulations on successfully installing / upgrading Jaeger v{version}!

📊 You can log into the Jaeger Query UI here:
```

charts/jaeger/Chart.yaml
danish9039: > another question - can a user override this version without having to upgrade the chart itself? I understand it may not work if there are breaking config changes, but it's still nice to give that flexibility _and document it_. They can always work around breaking config changes by supplying the full custom config.

yes , user can override version using flag `--set allInOne.image.tag=2.13.0` or via values file as well , i have documented this .

danish9039: switched to v2.12.0 , also tested the auto update workflow using dry run ,  `./.github/scripts/update-jaeger-version.sh --dry-run`  , its working .

charts/jaeger/templates/cassandra-schema-job.yaml
```diff
@@ -93,7 +92,7 @@ spec:
       {{- end }}
       {{- if .Values.storage.cassandra.tls.enabled }}
         - name: {{ .Values.storage.cassandra.tls.secretName }}
-          secret:
+           secret:
```
yurishkuro, line 95: whitespace issue?

Stevenpc3:
We will be moving to the Jaeger 2 chart version on the start of the new year.  Though it appears we will have a lot of work to do.  We currently have several jaeger instances deployed with a collector, query, and oauth.  We also use the cleaner and spark (the cleaner needs some Index lifecycle managment love for elasticsearch, but I see the move to ECK 9 so that is great).

We are not fully familiar with the archictecture yet, but from what I read a while ago it is a single binary that just takes a configurtaion, so from that, I would think a lot of the yaml files for deployments, services, and other resources would have remained and just been updated to accept a config and have a default one. It does not make sense to deploy the chart several times as listed in the readme when a chart can easily deploy many containers and configure them in one go... that is what makes charts so useful.

We also cannot move to the operator yet per external reasons, but hope to soon.  I assume the operator supports a microservice architecture.

We also wonder if you can have multiple collectors and query pods to support scaling.  We digest roughly 100 - 200 GB a day at times per instace with multiple instances (because last time we tried we could not have a single Jaeger support multiple namespaces and clusters per some UI trace identification issues that I think an issue was written for)  Jaeger assumes single instance.

I am sure I will have more insight when the work starts soon, but any answers ahead of then would be helpful.  Particularly why this chart could not support the microservice and it was all ripped out, as I said it likely would have as the same parts would be needed just a different configuration on start which can be done many ways.

Also, why remove oauth?  Does Jaeger now support oauth internally?

PR #696, opened 2025-12-14
[jaeger] Replace outdated bitnami charts

jkowall:
#### What this PR does
Remove bitnami charts which are dead and not updated
Added charts for Strimzi (Kafka) and Elastic official chart

Fixes security issues

#### Checklist

- [X] Chart Version bumped
- [X] README.md has been updated to match version/contain new values

### DIFF
--- a/run_ct_local.sh
+++ b/run_ct_local.sh
@@ -10,4 +10,6 @@ ct install --config ct.yaml \
     --set elasticsearch.data.replicaCount=0 \
     --set elasticsearch.coordinating.replicaCount=0 \
     --set elasticsearch.ingest.replicaCount=0 \
-    --set elasticsearch.clusterHealthCheckParams=wait_for_status=yellow&timeout=1s"
+    --set elasticsearch.clusterHealthCheckParams=wait_for_status=yellow&timeout=1s \
+    --set storage.elasticsearch.scheme=https \
+    --set storage.elasticsearch.tls.insecure=true"
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -70,5 +70,6 @@ jobs:
                 --set elasticsearch.data.replicaCount=0 \
                 --set elasticsearch.coordinating.replicaCount=0 \
                 --set elasticsearch.ingest.replicaCount=0 \
-                --set elasticsearch.clusterHealthCheckParams=wait_for_status=yellow&timeout=1s"
-
+                --set elasticsearch.clusterHealthCheckParams=wait_for_status=yellow&timeout=1s \
+                --set storage.elasticsearch.scheme=https \
+                --set storage.elasticsearch.tls.insecure=true"


jkowall:
> We don't support Kafka in the new chart, can remove it. And why not chart from confluence, main contributor?

Okay I will remove it. Strimzi is a CNCF project and is used to run Kafka on Kubernetes so I figured it was the right choice.

ct.yaml
```diff
@@ -6,5 +6,7 @@ chart-repos:
   - incubator=https://charts.helm.sh/incubator
   - elastic=https://helm.elastic.co
   - bitnami=https://charts.bitnami.com/bitnami
+validate-chart-schema: false
+validate-yaml: false
```
yurishkuro: Why would we not want to validate? It's just a recipe for bugs.

jkowall: Yeah sorry, was debugging but fixed it now.

charts/jaeger/templates/_helpers.tpl
```diff
@@ -156,7 +156,17 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
+{{- if $es.masterService }}
+{{- $host = $es.masterService }}
+{{- else if $es.fullnameOverride }}
+{{- $host = $es.fullnameOverride }}
+{{- else if $es.nameOverride }}
+{{- $host = printf "%s-master" $es.nameOverride }}
+{{- else -}}
+{{- $clusterName := default "elasticsearch" $es.clusterName }}
+{{- $host = printf "%s-master" $clusterName }}
+{{- end }}
```
yurishkuro, line 169: I really don't understand why we need all this complexity. I think it's none of Jaeger's business how users want to provision ES. Just look at this, a dozen files in our chart have to mention ES:
```
$ rg -l elastic
charts/jaeger/Chart.lock
charts/jaeger/Chart.yaml
charts/jaeger/README.md
ct.yaml
charts/jaeger/values.yaml
charts/jaeger/templates/elasticsearch-secret.yaml
charts/jaeger/templates/es-rollover-cronjob.yaml
charts/jaeger/templates/es-lookback-cronjob.yaml
charts/jaeger/templates/allinone-deploy.yaml
charts/jaeger/templates/es-index-cleaner-cronjob.yaml
charts/jaeger/templates/es-rollover-hook.yml
charts/jaeger/templates/spark-cronjob.yaml
charts/jaeger/templates/_helpers.tpl
charts/jaeger/templates/storage-configmap.yaml
```

I wouldn't even have any dependency on it in the Jaeger chart. But if we really-really want to make it very easy to setup Jaeger with ES via `--set provisionDataStore.elasticsearch=true`, fine, we have the default `elasticsearch:` settings in values.yaml, that is passed through to ES chart. If users want to customize those settings, they can edit that section. Why do we need any other logic? All the convoluted logic above is not tested in the CI, so I have no idea if it works.

This is not helping users, it's making them learn some completely arbitrary rules / DSL instead of relying on standard configs.

jkowall: Those were already there before; I can clean it up if you want. I didn't try to clean up all of those templates.

yurishkuro: I know, but this PR is now adding these 10 more lines of some obscure override logic. Are we trying to guess which name ES chart gives to its endpoint?

But to the larger point - yes I think we need to revisit the overall chart organization and remove all custom overrides logic. I think we need to keep it simple - if you want defaults, we provide them. If you want to customize - you provide full configuration.

jkowall: The challenge is that the Elastic official chart uses different settings than the older Bitnami one, so the goal is to make it work. If we want to fix the fundamental issues, we should revisit all that logic. I can give it a go if you'd like.

run_ct_local.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -e
+ct install --config ct.yaml \
```
yurishkuro, line 3: is this different from how lint-test workflow calls this? If not, I would prefer to remove this shell script, add a Makefile with a target `test` containing this exact command, and change the workflow to call `make test-es`

yurishkuro: https://github.com/jaegertracing/helm-charts/pull/698

PR #713, opened 2026-01-06
[jaeger] Fix broken template references in es-maintenance and rollover

### LINKED ISSUE #712: [Bug]: Template es-maintenance.yaml fails in Helm chart version 4.3.0
### What happened?

When installing Helm chart version 4.3.0 with esIndexCleaner enabled, an error occurs.
This is probably because the template elasticsearch.client.url has been removed, but it's still in comments in file jaeger/templates/es-maintenance.yaml. 

Example:
              # URL is now expected to be passed via args or env if needed, or configured in external scripts
              # {{ include "elasticsearch.client.url" . }} 

Because YAML comments are only evaluated after processing Go template expressions, Helm still wants to resolve the non existing template.

### Steps to reproduce

execute
`helm template my-jaeger jaegertracing/jaeger --version 4.3.0 --set esIndexCleaner.enabled=true`

### Expected behavior

Error shouldn't occur. Remove the comments or  escape them like # {{/* {{ include "elasticsearch.client.url" . }}  */}}

### Relevant log output

```shell
Error: jaeger/templates/es-maintenance.yaml:53:19
  executing "jaeger/templates/es-maintenance.yaml" at <include "elasticsearch.client.url" .>:
    error calling include:
template: no template "elasticsearch.client.url" associated with template "gotpl"
```

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

```shell

```

DCchoudhury15:
#### What this PR does
Removes broken references to the non-existent `elasticsearch.client.url` template helper.
- Removed the commented-out reference in `templates/es-maintenance.yaml` which was causing Helm template rendering to crash.
- Removed the active reference in `templates/es-rollover-hook.yml` which would have caused a crash if the rollover hook was enabled.
- Bumped chart version to `4.3.1`.

fixes #712

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -32,7 +32,6 @@ jobs:
             This issue has been automatically closed due to inactivity.
           stale-issue-label: 'stale'
           exempt-issue-labels: 'do-not-expire,help-wanted'
-          only-issue-labels: 'question'
           
           # Pull requests configuration  
           days-before-pr-stale: 60
--- a/charts/jaeger/README.md
+++ b/charts/jaeger/README.md
@@ -124,6 +124,18 @@ spark:
 
 For a full list of supported environment variables, see the [Spark Dependencies README](https://github.com/jaegertracing/spark-dependencies#readme).
 
+### 5. Query UI
+
+To enable the query ui, you need to enable the ingress and fill at least one host:
+
+```yaml
+jaeger:
+  ingress:
+    enabled: true
+    hosts:
+      - <fill a host here>
+```
+
 ## Configuring the Collector
 
 The Jaeger v2 configuration is defined in `config` using OpenTelemetry Collector syntax. You can override pipelines, receivers, and processors there.
--- a/charts/jaeger/templates/_helpers.tpl
+++ b/charts/jaeger/templates/_helpers.tpl
@@ -449,19 +449,6 @@ Create pull secrets for esLookback image
 {{- end }}
 
 
-{{/*
-Define curl image declaration
-*/}}
-{{- define "curl.image" -}}
-{{- $image := "curlimages/curl" -}}
-{{- if .Values.global.imageRegistry -}}
-{{ .Values.global.imageRegistry }}/{{ $image }}
-{{- else -}}
-{{ $image }}
-{{- end -}}
-{{- end -}}
-
-
 {{- define "jaeger.namespace" -}}
   {{- if .Values.namespaceOverride -}}
     {{- .Values.namespaceOverride -}}
--- a/charts/jaeger/templates/jaeger-ing.yaml
+++ b/charts/jaeger/templates/jaeger-ing.yaml
@@ -21,7 +21,7 @@ spec:
         paths:
           - path: /
             pathType: {{ default "ImplementationSpecific" $.Values.jaeger.ingress.pathType }}
-            backend: {{- include "common.ingress.backend" (dict "serviceName" (printf "%s-query" (include "jaeger.fullname" $)) "servicePort" 16686 "context" $) | nindent 14 }}
+            backend: {{- include "common.ingress.backend" (dict "serviceName" (include "jaeger.fullname" $) "servicePort" 16686 "context" $) | nindent 14 }}
   {{- end -}}
   {{- if .Values.jaeger.ingress.tls }}
   tls:
--- a/charts/jaeger/templates/ui-config.yaml
+++ b/charts/jaeger/templates/ui-config.yaml
@@ -1,0 +1,12 @@
+{{- if .Values.uiconfig -}}
+# Generates a config map from a file provided by user via `--set-file uiconfig=`
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ui-config
+  namespace: {{ include "jaeger.namespace" . }}
+  labels:
+    {{- include "jaeger.labels" . | nindent 4 }}
+data:
+  ui-config.json: {{ toJson .Values.uiconfig | quote }}
+{{- end -}}
--- a/charts/jaeger/templates/user-config.yaml
+++ b/charts/jaeger/templates/user-config.yaml
@@ -9,21 +9,5 @@ metadata:
     {{- include "jaeger.labels" . | nindent 4 }}
 data:
   user-config.yaml: |
-    {{- .Values.userconfig | nindent 4 }}
-{{- end -}}
-{{- if .Values.uiconfig -}}
-{{- if .Values.userconfig }}
----
-{{- end }}
-# Generates a config map from a file provided by user via `--set-file uiconfig=`
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: ui-config
-  namespace: {{ include "jaeger.namespace" . }}
-  labels:
-    {{- include "jaeger.labels" . | nindent 4 }}
-data:
-  ui-config.json: |
-    {{- .Values.uiconfig | nindent 4 }}
+{{ toYaml .Values.userconfig | indent 4 }}
 {{- end -}}
--- a/charts/jaeger/values.yaml
+++ b/charts/jaeger/values.yaml
@@ -2,9 +2,132 @@
 # This is a YAML-formatted file.
 # Jaeger values are grouped by component. Cassandra values override subchart values
 
-# Will use for user provided config
-userconfig:
+# User provided ui config: see https://www.jaegertracing.io/docs/latest/deployment/frontend-ui/#configuration for more examples and settings
+uiconfig:
+  # {
+  #   "dependencies": {
+  #     "dagMaxNumServices": 200,
+  #     "menuEnabled": true
+  #   },
+  #   "monitor": {
+  #     "menuEnabled": true
+  #   },
+  #   "themes": {
+  #     "enabled": true
+  #   },
+  #   "archiveEnabled": true,
+  #   "tracking": {
+  #     "gaID": "UA-000000-2",
+  #     "trackErrors": true
+  #   },
+  #   "menu": [
+  #     {
+  #       "label": "About Jaeger",
+  #       "items": [
+  #         {
+  #           "label": "GitHub",
+  #           "url": "https://github.com/jaegertracing/jaeger"
+  #         },
+  #         {
+  #           "label": "Docs",
+  #           "url": "https://www.jaegertracing.io/docs/latest/"
+  #         }
+  #       ]
+  #     }
+  #   ],
+  #   "search": {
+  #     "maxLookback": {
+  #       "label": "2 Days",
+  #       "value": "2d"
+  #     },
+  #     "maxLimit": 1500
+  #   },
+  #   "linkPatterns": [{
+  #     "type": "process",
+  #     "key": "jaeger.version",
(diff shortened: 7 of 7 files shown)


charts/jaeger/templates/es-maintenance.yaml
```diff
@@ -334,7 +331,7 @@ spec:
           {{- end }}
 {{- end -}}
 
-{{- if and .Values.esLookback.enabled .Values.esLookback.serviceAccount.create -}}
+{{ if and .Values.esLookback.enabled .Values.esLookback.serviceAccount.create }}
```
yurishkuro, line 334: why these changes?

Stevenpc3:
I came here to make an issue and PR for this very reason too.  When I run with the changes in es-maintenance.yaml I get the following template

``` yaml
---
# Source: jaeger/templates/es-maintenance.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: jaeger-es-index-cleaner
  labels:
    helm.sh/chart: jaeger-4.3.0
    app.kubernetes.io/name: jaeger
    app.kubernetes.io/instance: jaeger
    app.kubernetes.io/version: "2.14.1"
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/component: es-index-cleaner
automountServiceAccountToken: false
---
# Source: jaeger/templates/es-maintenance.yaml
apiVersion: batch/v1
kind: CronJob
metadata:
  name: jaeger-es-index-cleaner
  labels:
    helm.sh/chart: jaeger-4.3.0
    app.kubernetes.io/name: jaeger
    app.kubernetes.io/instance: jaeger
    app.kubernetes.io/version: "2.14.1"
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/component: es-index-cleaner
spec:
  concurrencyPolicy: Forbid
  schedule: "55 23 * * *"
  successfulJobsHistoryLimit: 3
  failedJobsHistoryLimit: 3
  suspend: false
  jobTemplate:
    spec:
      template:
        metadata:
          labels:
            app.kubernetes.io/name: jaeger
            app.kubernetes.io/instance: jaeger
            app.kubernetes.io/component: es-index-cleaner
        spec:
          serviceAccountName: jaeger-es-index-cleaner
          
          securityContext:
            runAsUser: 1000
          containers:
          - name: jaeger-es-index-cleaner
            securityContext:
              runAsUser: 1000
            image: jaegertracing/jaeger-es-index-cleaner:2.14.1
            imagePullPolicy: IfNotPresent
            args:
              - "7"
              # URL is now expected to be passed via args or env if needed, or configured in external scripts
              
              
            env:
              
            resources:
              {}
            volumeMounts:
          restartPolicy: OnFailure
          volumes:
```

which seems to deploy fine as you can see here

{C0B622A6-E811-44F6-A4E4-CCE96C918D91}

Stevenpc3:
The problem has to do with helper `elasticsearch.env` and when storage type is elasticsearch.  I am trying to figure it out now.

charts/jaeger/templates/es-maintenance.yaml
Stevenpc3: this is needed to stop the problem described below.  Without it you get this in the template generation

{7865C468-2335-4420-A163-4022CA0ADC74}

with the removal of those it will generate correctly.

DCchoudhury15: @Stevenpc3 yes same happened to me and when these changes were made it didnt throw this err anymore

charts/jaeger/templates/es-maintenance.yaml
```diff
@@ -96,9 +95,9 @@ spec:
               secret:
                 secretName: {{ .secretName }}
           {{- end }}
-{{- end -}}
+{{ end }}
```
Stevenpc3, line 98: add these back only the - around the if below is needed

charts/jaeger/templates/es-maintenance.yaml
Stevenpc3: There are 5 locations in this file that have the if and separator.  I would remove the - around the if on all of those lines.

charts/jaeger/templates/es-maintenance.yaml
```diff
@@ -299,7 +297,6 @@ spec:
             args:
               - lookback
               # URL must be passed via extraArgs or env
```
Stevenpc3, line 299: should just remove these comments too

Stevenpc3: which are not REAL comments anyway in here.  A real comment is 

```{{/* anything between here is a comment */}}``` 

which is why the include below is still being called.

**Comment by Stevenpc3** [comment]
I found a similar issue when enabling spark

{C32453E6-BA7E-4562-8906-223D6D09C7C6}

I think all the places that have this need the `-` removed in the if 

{9DFD0980-D649-43EA-B214-8D6E9CFB41A0}

this is in the spark-dependencies.yaml

**Comment by Stevenpc3** [comment]
Version need to bump to 4.3.3 because latest is 4.3.2 https://github.com/jaegertracing/helm-charts/blob/main/charts/jaeger/Chart.yaml#L6

File: charts/jaeger/templates/es-maintenance.yaml [hunk 4]
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.esIndexCleaner.enabled -}}
```
yurishkuro, line 1: What is the objective of removing `-`, which are used to suppress whitespace?

DCchoudhury15: Removing the whitespace chomping (`-`) prevents the `---` document separators from merging with adjacent lines during rendering. This fixes the `yaml: could not find expected ':'` error(which i was and @Stevenpc3 were facing ),  ensuring the `ServiceAccount` and `CronJob` are correctly separated in the generated output.

charts/jaeger/templates/es-maintenance.yaml
```diff
@@ -230,9 +228,9 @@ metadata:
     "helm.sh/hook-weight": "-10"
     "helm.sh/hook-delete-policy": before-hook-creation
 automountServiceAccountToken: {{ .Values.esRollover.serviceAccount.automountServiceAccountToken }}
-{{- end -}}
+{{ end }}
 
-{{- if .Values.esLookback.enabled -}}
+{{ if .Values.esLookback.enabled }}
```
yurishkuro, line 233: sorry, what is `esLookback`? V1 chart had no comments about it either.

DCchoudhury15: `esLookback` configures the **Elasticsearch Index Cleaner** CronJob (responsible for deleting old traces/indices).I did not introduce this variable; it was preexisting in the chart. I am only adjusting the whitespace logic around it to prevent the YAML parsing error (`could not find expected ':'`) that occurs when it is rendered.

Stevenpc3: I never enabled it in the past either. I only ever used esIdexCleaner and spark. 

This will also go away if Jaeger moved to data streams as I saw planned in another PR.

yurishkuro: @DCchoudhury15 the cleaner job has its own template. I understand cleaner and rollover, I don't understand what 'lookback' means
```
$ rg 'if .Values.*.enabled' charts/jaeger/templates/es-maintenance.yaml
1:{{- if .Values.esIndexCleaner.enabled -}}
117:{{- if .Values.esRollover.enabled -}}
235:{{- if .Values.esLookback.enabled -}}
```

DCchoudhury15: I didn't introduce the `esLookback` block—it was already present in the chart. Based on the arguments (`args: ["lookback"]`), it appears to create a separate CronJob for a specific "lookback" cleaning strategy, distinct from the standard `esIndexCleaner`.

My goal was just to fix the syntax error that causes the chart to crash when this block is used. However, if you believe this is dead code, I can remove the `esLookback` block entirely and verify that the `es-maintenance.yaml` template still generates correctly without it. Would you prefer I do that?

Stevenpc3: Funny the Google AI does a good job explaining it.

Elasticseach has the concept of ILM or index lifecycle management for data clean up that elasticsearch self manages when you create an ILM which is usually when the index is initially created (start of up Jaeger).  The lookback is a script that runs from jaeger.

I would not delete it.  At least not this time.  I know in the Jaeger repo there is active discussion about data streams which might replace the current indexes.  That would have a completely different index management cycle.  For now, these might both be used by others (other than the lookback hook hanging).

Also, I am not 100% sure the ILM even works correctly so the look back might be all that works right now (I will check later on my work computer).  I suggest we just make an issue to verify they are good.  They both use the same image to run.  We could also ask the original auther of those sections and we would check the code in the Jaeger repo to see what is actually in the image.

The rollover isn't even a "real" rollover either.  Since it's a job and jaeger indexes use a non standard index pattern that ends in date which is very old style and prevents elasticsearch ILM from actually working as intedended.  So BOTH "rollover" and "lookback" run as a cronjob where Jaeger does the management and not elasticsearch.

Here is the AI summary which is mostly correct.

image

Stevenpc3: These are the current settings we use (and expect to continue working)

``` yaml
    esIndexCleaner:
      enabled: true
      schedule: "55 23 * * *"
      resources:
        requests:
          cpu: 256m
          memory: 128Mi
      numberOfDays: 5
    spark:
      enabled: true
      schedule: "00 21 * * *"
      resources:
        requests:
          cpu: 256m
          memory: 4G
```
as of ```jaeger-es-index-cleaner:1.71.0``` here is the output log output which we run multiple times a day (thought only needs to run once a day)

{C4E756BE-2440-42C3-84F1-0FB09451E180}

Stevenpc3: actually why there is cleaner, rollover, lookback is a bit confusing.  I know we only use the cleaner.  I can dig more into what it is this week as I work on it and if I figure it out I can try to add documentation somewhere.  Whatever it is, Jeager images manage the indexes in all cases instead of elasticsearch.

Stevenpc3: a quick idea (based on slowly becoming more awake)

cleaner - just runs and index curl and then removes items based on date.

rollover - is likely used to set a DIFFERNET ILM than the one created by default which is "per day" as that is potentially a very large index (we get like 200GB a day on some indexes) and that is poor performance.  Elasticsearch should have 50GB index max for performance which means when it hits a certain size or time it "rolls over" and puts that index into warm or cold and creates a new hot index (like partitions in sql dbs) and we should be using that actually...

lookback - seems like it does the same as cleaner lol.  I am not sure... this is a lot of typing to come to the same question of "what _does_ it do?" It seems to be another for of "cleaner" based on this https://github.com/jaegertracing/jaeger/issues/3368 which the helper outputs ```used with lookback to remove indices from read alias e.g, days, weeks, months... ``` for the ```unit```

Actually, the jaeger docs explain it well.  Lookback is just cleaner essentially.  You can use either https://www.jaegertracing.io/docs/2.14/storage/elasticsearch/#roll-over

So it should stay.

PR #718, opened 2026-01-09
[jaeger] Fix backend service name in jaeger-ing.yaml

ralphg6:
#### What this PR does

It fixes the relationship between the ingress and the service, allowing access to the Query UI as well.

It includes short instructions on how to enable the query UI.

fixes https://github.com/jaegertracing/helm-charts/issues/719

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/README.md
+++ b/charts/jaeger/README.md
@@ -133,7 +133,7 @@ jaeger:
   ingress:
     enabled: true
     hosts:
-      - myhost.example.com
+      - <fill a host here>
 ```
 
 ## Configuring the Collector


charts/jaeger/README.md
```diff
@@ -124,6 +124,18 @@ spark:
...
+### 5. Query UI
+
+To enable the query ui, you need to enable the ingress and fill at least one host:
+
+```yaml
+jaeger:
+  ingress:
+    enabled: true
+    hosts:
+      - myhost.example.com
```
jaegertracingbot: where is the host name supposed to come from?

ralphg6: We need to fill it with a host that reaches the ingress controller, as is usual when using ingresses on k8s.

yurishkuro (approved):

Issue #720, opened 2026-01-10
[Bug]: userconfig does not work


Stevenpc3:
### What happened?

I want to create a ```userconfig``` to override the config generated via ```storage-configmap.yaml``` but the ```user-config.yaml``` is broken.

It current produces this error when you set anything under ```userconfig``` in the values.yaml

 
<img width="2143" height="61" alt="Image" src="https://github.com/user-attachments/assets/2d975276-e810-4488-b765-dcc3770cc525" />

### Steps to reproduce

place something under userconfig in the values.yaml

template or install

### Expected behavior

user-config.yaml is generated and no error is produced

### Relevant log output

```shell
Error: template: jaeger/templates/user-config.yaml:12:37: executing "jaeger/templates/user-config.yaml" at <4>: wrong type for value; expected string; got map[string]interface {}
```

### Screenshot

<img width="2245" height="155" alt="Image" src="https://github.com/user-attachments/assets/0cff0f61-36fc-4d70-ad10-bf1362fbf53e" />

### Jaeger backend version

2.x

### SDK

NA

### Pipeline

NA

### Stogage backend

NA

### Operating system

NA

### Deployment model

NA

### Deployment configs

```shell

```

PR #721, opened 2026-01-10
Fix user-config.yaml

Stevenpc3:
#### What this PR does
Fixes broken user-config.yaml template

https://github.com/jaegertracing/helm-charts/issues/720

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/values.yaml
+++ b/charts/jaeger/values.yaml
@@ -177,7 +177,7 @@ config:
             max_traces: 100000
         archive_store:
           elasticsearch:
-            # index_prefix: jaeger
+            # index_prefix: jaeger-archive
             server_urls: ["http://elasticsearch-master:9200"]
             # auth:
             #   basic:


charts/jaeger/templates/user-config.yaml
```diff
@@ -25,5 +25,5 @@ metadata:
     {{- include "jaeger.labels" . | nindent 4 }}
 data:
   ui-config.json: |
-    {{- .Values.uiconfig | nindent 4 }}
+{{ toYaml .Values.uiconfig | indent 4 }}
```
yurishkuro: Why "to yaml" for json file?

Stevenpc3: Good catch.  I honeslty just made an assumption after it generated correctly.  I did not notice that it was a json or what the ui-config was.  I see it here though https://www.jaegertracing.io/docs/2.dev/deployment/frontend-ui/ I will cahnge it to toJson and test it generates correctly.

Stevenpc3: fixed

charts/jaeger/values.yaml
```diff
@@ -46,16 +169,20 @@ config:
...
             max_traces: 100000
         archive_store:
           elasticsearch:
-            # index_prefix: jaeger-archive
+            # index_prefix: jaeger
             server_urls: ["http://elasticsearch-master:9200"]
-            # username: elastic
-            # password: changeme
+            auth:
+              basic:
```
Stevenpc3: this is to be commented out

charts/jaeger/templates/user-config.yaml
```diff
@@ -9,21 +9,5 @@ metadata:
...
-kind: ConfigMap
-metadata:
-  name: ui-config
-  namespace: {{ include "jaeger.namespace" . }}
-  labels:
-    {{- include "jaeger.labels" . | nindent 4 }}
-data:
-  ui-config.json: |
-    {{- .Values.uiconfig | nindent 4 }}
+{{ toYaml .Values.userconfig | indent 4 }}
```
yurishkuro, line 12: I don't believe this was the intention of either `userconfig` or `uiconfig` fields. The toYaml/toJson commands take _the value_ from values.yaml and re-render that value into output format. There is no point having `userconfig` that behaves this way if we already have `config` that is also fully user-editable and is the same by-value thing. The point of userconfig and uiconfig was to point to _file names_ with the configs, and use something like `{{ .Files.Get .Values.uiconfig }}`

As I mentioned, I don't have much experience with Helm so maybe passing external files is non-idiomatic and we can abandon this way of customization in favor of how you implemented it, but in this case I'd argue we can drop the `userconfig` altogether. Or we can do `config` and `uiconfig` for by-value instantiation and `config_file` / `uiconfig_file` for external file (which could be used with schema validation or 3rd party tools that generate Collector configs, like OTelBin).

Stevenpc3: I need to take this line by line so forgive me.

I am not sure what the intention was for the userconfig or uiconfig but _as written_ it takes the value as you stated.  This is normal and acceptable in other charts.  Or have "extraconfigs" for example or a configmap override. It is a bit confusing here because I too thought that config would be overridable, but the issue with that is you need your defaults to be correct and minimal and in this case they are not also there is this problem that prevents proper overriding currently in some cases https://github.com/jaegertracing/helm-charts/issues/722 .

For pointing to files, I WAS going to move uiconfig into a file, but that does not work for people that want to use the chart directly as they cannot edit or add a file to a chart.  I currently use the chart as a "wrapper" where I use jaeger as a dependency.  In that case I could simply make a configmap of my own and point userconfig to that configmap... except is not set up that way AT ALL for this to work currently. It would need to accept a name of a configmap then load that in and the names are also not good as they can collide with other values since they are coded and also no values are renderable so it cannot dynamically generate them. And some things like the secret file name and key used for elasticsearch external were removed with now requires hard coding passwords into the values.yaml or command line which was just a pointer to a secret in the past.

I have a list of fixes I want to get in and I agree with much of what you say as 4.1 or other minor additions, but for a bug fix this current PR and the other one https://github.com/jaegertracing/helm-charts/pull/713 the fixes need to go in just to make it fully usable.

To be honest, there are a large number of CVEs in Jaeger 1.x and I NEED to go to 2, but I cannot as the charts need emergency fixes.  After those go in I can make these features and other additions we think would be beneficial and clean it up more.  I am aiming for usable at the moment.

Stevenpc3: But yes I can override the primary store via something like 

``` shell
helm template jaeger charts/jaeger --set config.extensions.jaeger_storage.backends.primary_store=test
```

which produces  this 

{834AF10F-CE4A-439F-9306-4110397919B8}

but if I can't set just the elasticsearch fields without setting ```memory: null``` as helm will only overwrite values in a list and not remove them when joining overrides with the parent default values.yaml, so as above works because it replaces the ENTIRE primary_store object it fails when I use an override file like this which excludes the memory field

{97DA3292-3C95-4A26-B023-A723C0E65BAF}

The default config value is this which has the memory section

{54034999-DE84-4735-82C0-EC9999506E6E}

This command to apply the override file 

``` shell
helm template jaeger charts/jaeger -f override --output-dir out
```

produces this output section where you can see "memory" is not removed (which results in in memory storage NOT elasticsearch storage which is not desired)

{A156FFEE-4446-404B-AEEC-C54BF1B8F741}

So, while you CAN override some of the config you cannot reliably override it all without a userconfig or pointing to something that will skip the usage of the "config"

The best solution for that is to do as it is here with userconfig that takes precedence over config or add a field like "config_configmap" which the name is bad (because ```config``` is a horrible name to begin with) and anything with underscore like jaeger_storage for example also requires the use of ```{{ index .Values.config.extensions "jaeger_storage".backends "archive_store".elasticsearch.auth.blah }}``` because an underscore is not valid in helm otherwise.  So underscore should be avoided.

Removing userconfig or uiconfig now would also be a breaking change and require moving to 5.x.y of the charts. So would changing the names.

DCchoudhury15: @Stevenpc3 Thanks for handling the refactor to split the files! Since you've implemented the fix (and the file separation) in your changes, should i close #713  in favor of this solution to avoid duplication, or is it important for the fix.

Stevenpc3: I did not split the items in your PR into files.  I split a different set of files. 

I CAN do that here if that would simplify things.  Though I will not be able to do it until later this even for me.

Stevenpc3: I prefer to keep PRs clean and separate though.  I can make a commit to your PR also if that would be better.

DCchoudhury15: I saw you opened #721 for the user-config fixes, so I will focus on cleaning up (#713).
I just want to make sure I am understanding the next steps correctly before I push changes:

For es-maintenance.yaml: I know you mentioned splitting this into separate files to avoid the separator complexity. Is it okay if I just fix the formatting issues for now (by removing the dashes - from the if statements) to get this merged quickly? Or is there another way you wanna go which i missed?

For es-rollover-hook.yml: Since we both encountered that hang/timeout issue when it's enabled, I am planning to revert the changes to this file completely so it doesn't block the maintenance fix.

Does that plan sound right to you, or am i missing something?

Stevenpc3: I think that is fine. I would remove the dashes in all the files including spark then template to ensure they are generated (extra newlines are fine).

For es-rollover-hook.yaml I guess revert for now then make an issue for that. This will leave it in a state of broken either way, but the current form of broken is certainly less disruptive.

charts/jaeger/values.yaml
```diff
@@ -2,9 +2,132 @@
 # This is a YAML-formatted file.
 # Jaeger values are grouped by component. Cassandra values override subchart values
 
-# Will use for user provided config
+# User provided ui config: see https://www.jaegertracing.io/docs/2.dev/deployment/frontend-ui/#configuration for more examples and settings
```
yurishkuro:
```suggestion
# User provided ui config: see https://www.jaegertracing.io/docs/latest/deployment/frontend-ui/#configuration for more examples and settings
```

charts/jaeger/values.yaml
```diff
@@ -2,9 +2,132 @@
...
+uiconfig:
+  # {
```
yurishkuro, line 7: JSON inside yaml is not valid, is it? The ui config will have to be formatted as yaml to be included like this, and toJson would translate that.

charts/jaeger/templates/user-config.yaml
yurishkuro: > but if I can't set just the elasticsearch fields without setting memory: null as helm will only overwrite values in a list and not remove them when joining overrides with the parent default values.yaml

btw having memory storage hanging in the config is not damaging if nothing actually uses it

charts/jaeger/values.yaml
Stevenpc3: Json is valid inside yaml https://helm.sh/docs/howto/charts_tips_and_tricks/#yaml-is-a-superset-of-json

This generates stringified json inside a configmap. That is parsed into correct json.

I am not 100% sure if that string json is read by Jaeger correctly, but Jaeger does at least start with the example uncommented. 

I can alter the ui settings and see if the UI changes later tonight. At this point I was aiming for the value to generate correctly as it did not before.

charts/jaeger/templates/user-config.yaml
Stevenpc3: That is my point... It IS used. This is passed directly to Jaeger and Jaeger will boot in memory storage mode if it is present. My other issue mentions this. With memory there, you can't use elastcsearch.

I found this while testing. And I can't currently use the chart specifically because of this issue. 

If Jaeger accepted it correctly as you mention them I would not have done this work right now. 

It would be good to confirm this as I laid out because I assume it's a Jaeger bug with precedence.

Stevenpc3: This is described above in detail and mentioned here https://github.com/jaegertracing/helm-charts/issues/722

Stevenpc3: The reason I found this was I deployed then set to use external elastic but it was resetting traces when the pod deleted and not creating indexes (after I fixed the password issue and did another work around to set elastic password because it's generated and I have a chart that deployed OUR elastic and Jaeger at the same time and can no longer point to a secret like the old chart) 

So I realized memory mode was set. To fix this I set memory to null to remove it from the config and it worked. Fine I said... But then we use Argo and there is a bug where the null trick does not work...

So I found the user config and noticed in the templates it overrides config which is cool and useful... But it is broken and throws errors. And I decided to fix ui config while there since it had the same errors. 

So, long road, but several issues identified.

charts/jaeger/values.yaml
```diff
@@ -46,16 +169,20 @@ config:
...
+            # auth:
+            #   basic:
+            #     username: admin
+            #     password: password
           memory:
             max_traces: 100000
         archive_store:
           elasticsearch:
-            # index_prefix: jaeger-archive
+            # index_prefix: jaeger
```
yurishkuro:
```suggestion
            # index_prefix: jaeger-archive
```

PR #724, opened 2026-01-14
[jaeger] Remove storage provisioning capabilities

### LINKED ISSUE #725: [Bug]: cannot run esIndexCleaner and other es maintenance jobs
### What happened?

When trying to run es maintenance jobs they fail because they require urls and other configurations.

### Steps to reproduce

enable and run the jobs

### Expected behavior

job runs without error

### Relevant log output

```shell
Error: wrong number of arguments
Usage:
  jaeger-es-index-cleaner NUM_OF_DAYS http://HOSTNAME:PORT [flags]

Flags:
      --archive                       Whether to remove archive indices. It works only for rollover
      --es.password string            The password required by storage
      --es.tls.ca string              Path to a TLS CA (Certification Authority) file used to verify the remote server(s) (by default will use the system truststore)
      --es.tls.cert string            Path to a TLS Certificate file, used to identify this process to the remote server(s)
      --es.tls.enabled                Enable TLS when talking to the remote server(s)
      --es.tls.key string             Path to a TLS Private Key file, used to identify this process to the remote server(s)
      --es.tls.server-name string     Override the TLS server name we expect in the certificate of the remote server(s)
      --es.tls.skip-host-verify       (insecure) Skip server's certificate chain and host name verification
      --es.username string            The username required by storage
      --feature-gates flag            Comma-delimited list of feature gate identifiers. Prefix with '-' to disable the feature. '+' or no prefix will enable the feature. (default configoptional.AddEnabledField,-confmap.enableMergeAppendOption,-es.index.relativeTimeIndexDeletion)
  -h, --help                          help for jaeger-es-index-cleaner
      --index-date-separator string   Index date separator (default "-")
      --index-prefix string           Index prefix
      --rollover                      Whether to remove indices created by rollover
      --timeout int                   Number of seconds to wait for master node response (default 120)

2026/01/14 20:53:07 wrong number of arguments
```

### Screenshot

Image

### Additional context

_No response_

### Jaeger backend version

2.x

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

elasticsearch

### Operating system

_No response_

### Deployment model

kubernetes and helm

### Deployment configs

```shell
helm upgrade -i jaeger charts/jaeger --set esIndexCleaner.enabled=true
```

Stevenpc3:
#### What this PR does

* fixes various templates and values.yaml settings to allow users to use external elasticsearch
* remove automatic storage provisioning

#### Fixes

Resolves #725

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 2.14.1
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
-version: 4.3.5
+version: 4.4.0
 # CronJobs require v1.21
 kubeVersion: ">= 1.21-0"
 keywords:


charts/jaeger/values.yaml
```diff
@@ -163,26 +163,28 @@ config:
         traces: primary_store
         traces_archive: archive_store
 
+    # Adding memory to ANY storage will cause jeager to use memory.  To completely override and NOT use memory please see 'userconfig' above
+    # This chart defaults to using memory
```
yurishkuro: Why document incorrect behavior instead of fixing it? As I mentioned, Jaeger can have any number of storage backends configured, but it will only use those that are named in the storage & query extensions:
```
      storage:
        traces: primary_store
        traces_archive: archive_store
```

yurishkuro: @danish9039 do you know the background on why memory storage is overriding others?

Stevenpc3: I documented it because it's a current issue others need to be aware of. 

I also need this now, not later. And this is an issue with Jaeger not the charts. 

In this case "primary_storage" had memory AND elastcsearch defined and it defaults to memory of the two. 

I do not know if it is even acceptable to have multiple items listed there under ```primary_storage``` for example, but it's not prevented and that is how the example was. 

I suppose the better solution would be to define

primary_storage_memory
primary_storage_elastcsearch
archive_storage_memory 
archive_storage_elasticsearch

Then set something like when switching.

```
      storage:
        traces: primary_store_elasticsearch
        traces_archive: archive_store_elastcsearch
```
I can do that. Then I wouldn't need the ```memory: null``` trick or to use ```userconfig```

I will try that in the morning.

Stevenpc3:
```
jaeger_storage:
      backends:
        primary_store:
          elasticsearch:
            # index_prefix: jaeger
            server_urls: ["http://elasticsearch-master:9200"]
            # auth:
            #   basic:
            #     username: admin
            #     password: password
          memory:
            max_traces: 100000
```

Is technically valid and was the original example. This defaults to memory. I see your point now that this should not even be like this. It should be broken up into two stores. Then referenced as needed. 

It's this is what you mean then I get it and I didn't earlier.

Like this

```
jaeger_storage:
      backends:
        primary_store_elasticsearch
          elasticsearch:
            # index_prefix: jaeger
            server_urls: ["http://elasticsearch-master:9200"]
            # auth:
            #   basic:
            #     username: admin
            #     password: password
        primary_memory
          memory:
            max_traces: 100000

storage:
   traces: primary_store_elasticsearch
```

charts/jaeger/values.yaml
```diff
@@ -531,6 +545,7 @@ esRollover:
     #   memory: 128Mi
   serviceAccount:
     create: true
+    annotations: {}
```
Stevenpc3, line 397: This needs removed or added to the template

charts/jaeger/values.yaml
Stevenpc3: If we agree then I'll close the other issue. I would look into maybe enforcing that a defined store only contains one object?

yurishkuro: how do you mean "enforcing"?

the structure of the config is
```
backends:
  arbitrary_name_1:
    # one of (already enforced)
    memory:
    elasticsearch:
    ...
  arbitrary_name_2:
    # one of
    memory:
    elasticsearch:
    ...
```
It is by design that you can define multiple backends with different names and reference them from where needed. This allows, for example, to combine primary storage with one type and archive storage with another.

Stevenpc3: We seem to be talking past each other.

Let's just work off your example above and try to get on the same page.

In my testing, if I use your example and I set exporters and extentions.jaeger_query to "arbitrary_name_1" like this

``` yaml
config:
  extensions:
    jaeger_query:
      storage:
        traces: arbitrary_name_1
  exporters:
    jaeger_storage_exporter:
      trace_storage: arbitrary_name_1
```

Then Jager will have the OPTION of using ```memory``` or ```elasticsearch```.  I am stating it will ALWAYS choose memory.

How do you expect it to work in the case where it has to choose?  Would it use both?  How does it choose?  What defines which option of the 

{5670E6D3-92FC-435D-8832-13EFA4EF7C7F}

The closest I found that might be the case is a value in the jaeger-deploy.yaml called ```SPAN_STORAGE_TYPE```

{486B6A05-1AE1-4FC9-96D0-6B59C5DA295B}

Which is not used anywhere in the jaeger code other than reading it in.  Though it appears that is SUPPOSED to be the way it selects which to use https://github.com/search?q=repo%3Ajaegertracing%2Fjaeger%20SPAN_STORAGE_TYPE&type=code

```Storage``` is used https://github.com/search?q=repo%3Ajaegertracing%2Fjaeger+STORAGE&type=code&p=1

Also, in the examples here https://github.com/jaegertracing/jaeger/tree/main/cmd/jaeger there  is none with ```multiple storage types defined under a single arbitrary_name```

So can you point to anything that allows, shows, enforces having multiple stores under one name?  Otherwise, I think something is broken OR the expectation is that there is a single store defined under a single arbitrary_name.

I will test with the "STORAGE" key to see if that work instead of "SPAN_STORAGE_TYPE"  and then I will ALSO put a single store under the arbitrary_name as in the examples.

Then I will explain in the readme better (once we settle on the answer here) how to switch between, because with charts people usually expect to set a value to change configurations.  Here we default to memory so they would need to know how to change to NOT memory.

I think what is going on here is a mechanism is not working as expected and the example is one that is not used anywhere else, so it exposed a bug.

Comment (Stevenpc3) [reply charts/jaeger/values.yaml]: and perhaps you think there is enforcement in jaeger for this config where you cannot put multiple stores in a single backend definition?

{1613AD0D-1089-415E-AC9A-017BAB3D34EF}

it is not.  It will load just fine and will default to memory.  I have not tried with say elasticsearch and cassandra, but with elasticsearch and memory it uses memory and does not complain.

Comment (Stevenpc3) [reply charts/jaeger/values.yaml]: I have fixed the MR to be more correct.  I am fine with moving on.  I do think there might be an issue, but that happens when outside the norm.

File: charts/jaeger/values.yaml [hunk 3]
```diff
@@ -285,13 +293,6 @@ jaeger:
     #     hosts:
     #       - chart-example.local
     pathType:
-  # resources:
```
Stevenpc3, line 288: add this back

charts/jaeger/values.yaml
yurishkuro: you were right, we were missing validation for multiple types defined under a single backend name. Fixing in https://github.com/jaegertracing/jaeger/pull/7875

What that means for Helm chart is that if it creates invalid configs where more than one storage type is defined under a single backend name the binary won't start.

Stevenpc3: I fixed it to remove the multiple now and I think it reflects what is expected and is actually easier now.

Stevenpc3: so I think this MR is good to go now.

Makefile
```diff
@@ -9,20 +9,39 @@ lint:
 test:
 	ct install --config ct.yaml
 
-.PHONY: test-es
-test-es:
+# This test clears the extraEnvs for elasticsearch provision and then runs the elasticsearch builtin tests
+# jaeger comes up in memory storage mode
+# .PHONY: test-es
```
yurishkuro: should we delete this?

Makefile
```diff
@@ -9,20 +9,39 @@ lint:
...
+		--set elasticsearch.tests.enabled=false \
 		--set elasticsearch.readinessProbe.initialDelaySeconds=60 \
 		--set elasticsearch.readinessProbe.failureThreshold=10 \
-		--set storage.elasticsearch.scheme=https \
-		--set storage.elasticsearch.tls.insecure=true"
+		--set elasticsearch.volumeClaimTemplate.resources.requests.storage=5Gi \
+		--set elasticsearch.protocol=http \
+		--set config.extensions.jaeger_query.storage.traces=primary_store_elasticsearch \
+		--set config.extensions.jaeger_query.storage.traces_archive=archive_store_elasticsearch \
+		--set config.exporters.jaeger_storage_exporter.trace_storage=primary_store_elasticsearch "
```
yurishkuro: any thoughts one also enabling all those maintenance & spark jobs as part of this test?

Makefile
Stevenpc3: I can.  It is in the history.  I kept it for reference, but no real need to keep it.

Makefile
Stevenpc3: We can consider that for another follow on if you would like.  We can make an issue and I can try.  The issue would be setting up calls to the job or making cronjob run correctly.  That would be tough.

Also, spark is not working so I would skip that anyway.  And the only one that would make sense right now would be esIndexCleaner.

Makefile
Stevenpc3: removed

charts/jaeger/values.yaml
```diff
@@ -160,31 +161,36 @@ config:
...
-            # index_prefix: jaeger-archive
-            server_urls: ["http://elasticsearch-master:9200"]
-            # auth:
-            #   basic:
-            #     username: admin
-            #     password: password
+        archive_store_memory:
           memory:
             max_traces: 100000
+        primary_store_elasticsearch:
```
yurishkuro: is this section always going to be present in the final configmap? The current implementation of the backend factory unfortunately tries to initialize all storage factories on start-up, regardless of where the storage is referenced by anything in the config. That means that if there is no Elasticsearch running, Jaeger fails to start even if nothing actually needs ES storage. E.g. I just tried with this config:
```yaml
  jaeger_query:
    storage:
      traces: some_store
      traces_archive: another_store

  jaeger_storage:
    backends:
      some_store:
        memory:
          max_traces: 100000
      another_store:
        memory:
          max_traces: 100000
      elasticsearch:
        elasticsearch:
          server_urls:
            - http://localhost:9200
```
Log:
```
Error: failed to start extensions: failed to initialize storage 'elasticsearch': failed to create Elasticsearch client: health check timeout: Head "http://localhost:9200": dial tcp [::1]:9200: connect: connection refused: no Elasticsearch node available
2026/01/16 13:12:57 failed to start extensions: failed to initialize storage 'elasticsearch': failed to create Elasticsearch client: health check timeout: Head "http://localhost:9200": dial tcp [::1]:9200: connect: connection refused: no Elasticsearch node available
exit status 1
```

Stevenpc3: well lol crap.

So that seems like jaeger should not do that?  do it on demand when used?

I would have to remove it from the config, then I would have to make the configmap a template that is renderable and then make a helper to create the backend on demand based on configuration for memory or elasticsearch.  This would be to support the user just getting "memory" out of the box.  The idea being that memory is default mode and a toggle of a key value would put it in another mode.

Stevenpc3: this seems like it would be better to fix in jaeger to only load the ones that are referenced.

If that is not feasible then I can try something in the charts.

at least you found it now

yurishkuro: it's feasible, I am preparing ADR to see what it would take.

But I wonder why [test-with-all-in-one](https://github.com/jaegertracing/helm-charts/actions/runs/21074058384/job/60611220340?pr=724#logs) succeeded when ES storage was configured but not deployed.

Stevenpc3: does this fail in the new code if you deploy a similar example without using the helm charts?

I can look in the code to see if I can find why that might happen and maybe just point to the latest 2.14? jaeger and test it myself too.

Stevenpc3: I am testing with 2.14.1 -- are you using an unreleased version to test?

Stevenpc3: I assume it has to do with this 

{DF605566-BC80-4682-8FF1-C3D41AB8DE84}

the all in one does not pass in the config file.

Stevenpc3: which... is actually a bug in the helm chart in that you could not set any memory settings as the config would not get passed.

Stevenpc3: https://github.com/jaegertracing/helm-charts/blob/main/charts/jaeger/templates/storage-configmap.yaml loads the "config" key if custom "userconfig" is not set

https://github.com/jaegertracing/helm-charts/blob/main/charts/jaeger/templates/user-config.yaml loads the "userconfig" if userconfig is set

So, either way, a "user-config.yaml" file is created as a configmap.

but, jaeger-deploy.yaml will ONLY pass in the arg to load the "user-config.yaml" file if cassandra or elasticsearch or userconfig key are set (excluding any memory or default configurations) https://github.com/jaegertracing/helm-charts/blob/main/charts/jaeger/templates/jaeger-deploy.yaml#L65

it also only mounts the config in the same cases as above https://github.com/jaegertracing/helm-charts/blob/main/charts/jaeger/templates/jaeger-deploy.yaml#L120  and  https://github.com/jaegertracing/helm-charts/blob/main/charts/jaeger/templates/jaeger-deploy.yaml#L146

This should be changed to always mount and load the user-config.yaml file since it is created and used in the case of memory settings.

I can update that and I assume it might also fail then in 2.14.1

yurishkuro: I was running in unreleased code but there were no changes since last release to that part. 

> the all in one does not pass in the config file.

yes, most likely - all-in-one uses built-in config file. You can still affect settings in this file using `--set` flags to the binary, but I doubt the chart does that.

Stevenpc3: I made the changes to ensure the config is passed as expected in the current version with the memory setting and it FAILS as you stated.

{BAAB9A7E-A9C0-45C2-95BF-9D9DB92477C2}

yurishkuro: https://github.com/jaegertracing/jaeger/blob/main/docs/adr/003-lazy-storage-factory-initialization.md

Stevenpc3: I do think there is a bug in the chart as configured... but fixing that bug will break the chart lol.

I am not 100% sure on the best solution here as of right now other than Jaeger allowing it.  I will ahve to think, but there are ways to ensure this is good in the chart at least for default and provisioned settings.

Stevenpc3: the chart should pass in the config even for the all in one otherwise, the --set flags to override settings would not work.

yurishkuro: I was referring to `--set` flags passed to Jaeger binary, not to Helm chart.
```
$ go run ./cmd/jaeger \
  --set extensions.jaeger_storage.backends.some_storage.memory.max_traces=123
...
Initializing in-memory storage with max_traces=123
```

Stevenpc3: Oh ok makes sense. 

I still need this MR to go in and I assume that ADR will be a while, so I'll try to make some more fixes for the chart which needs the config passed in even for memory anyway and currently it's not.

Did you write that ADR? It's very detailed and that was very quick. It also seems correct with the determinations.

Stevenpc3: it's interesting that helm and jaeger have the same --set structure.  That could be exploited to make the helm charts much more simple, but that is not in scope for this PR.  And I assume would require modifications to the Dockerfile.

Stevenpc3: the problem I am seeing with the chart now is that the config does not support cassandra currently, so that should be added at some point, but without the ADR fix, that would mean dynamically generating or modifying the config in a way that allows users to override the values as they can't pass a --set override to something that doesn't exist.

Then there is the possibility to force users to use either in memory if if you want something else then you have to pass in a userconfig as json or allow it to point to a configmap they create.  Which this is also fine, but then if they want to use memory we NEED the default config again to allow overrides.  Then if they do override the config with something other than memory settings it will break for reasons in the ADR.

So, without the ADR solution, this is a ticking time bomb for the chart or a very complicated bunch of validation/generation logic that is better left to Jaeger.

Stevenpc3: we could just go default and ANY custom config requires a userconfig... even memory modifications.  That would be a little annoying to users, but they would quickly get over it and then it would be simple to have a list of your own custom configmaps that you just pass in as an override with a -f memory.yaml or -f elastitcInternal.yaml or -f elasticExternal.yaml etc

That is the simplest and most future proof approach.  And if laid out this way in teh Readme it would be pretty clear and clean.

Stevenpc3: I will push that up for now...

Stevenpc3: actually.. no I won't.  The issue is the internal provisioned elaticsearch and cassandra need configurations which cassandra doesn't have so I am not even sure that works with the chart as it has no connection details for jaeger.  This chart only really currently work out of the box in memory and thre rest is an exercise of discovery for the user.

it seems like it would be the easiest to DROP the provisioned elastic and cassandra and just provide ample examples of how to configure them in the readme and leave that as a thing users need to set up.  (that is what we do for elasticsearch anyway)

Also, the current elasticserach for provisioning is VREY old and is the abandoned archived version from 3 years ago.  It is verison 8 and not great since jaeger supports 9.  The current Elastic is actually ECK which requires an operator to install and would not work in this chart and be waaaaaaaaaay out of scope and insane to manage (we have a chart already that does it... lots of work)

I think dropping everything other than Jaeger and asking users to provide configs as a custom setting is the right path for this chart.  (keep the es maintenance stuff and spark)

yurishkuro: > Did you write that ADR?

with Claude Code

yurishkuro: > the problem I am seeing with the chart now is that the config does not support cassandra currently

> it seems like it would be the easiest to DROP the provisioned elastic and cassandra and just provide ample examples

These are the points I've been raising repeatedly - I think the chart should be dead-simple and complex configuration issues solved by the user providing fully specified config. I really don't see any point in having `.Values.config` - it's some hardcoded configuration, with all kinds of issues (as you're discovering), it requires a bunch of logic in the templates to pull pieces of it. Why? What do we get from this, when the user (who is presumably familiar with OTEL Collector configuration approach) can just provide their own fully specified configuration (e.g. via userconfig)? The worst part to me about having `.Values.config` is that while it looks like OTEL config it's actually a custom DSL that the user needs to learn. Just let them learn OTEL config only.

I have less strong feelings about provisioning storage. As a point of minimizing the chart complexity I am 100% on board. But if the chart was able to support switching from memory to Elasticsearch-based deployment with a single flag, then I think it's a good user experience. Having said that, if self-provisioning would involve just another helm command and then using an example config that we provide, that would be almost as good user experience but will remove huge complexity from our chart.

Stevenpc3: Yeah we have a "wrapper" chart we deploy.  Our deps are jaeger, our own eck-stack-wrapper (we have the operator deployed already), and our own oauth chart.  This gives us everything. 

Keeping provisioning is part of the problem.  A flick of the flag to deploy provisioned elastic (again out of date and not maintained chart so it's EOL anyway) would require this chart to provide a config by default which it can or the user to provide a config for the provisioned one.  I think that is silly.

I am adding to this MR a commit that rips it all out then we can discuss.  There are a few values passed in as an ENV var to jaeger that are very old... I am curious if they are even required anymore.

yurishkuro: Jaeger v2 does not support any env vars internally anymore. A user can write the config file that references env vars (can be seen in some of our sample configs in the repo), but that has nothing to do with Jaeger logic.

Stevenpc3: so you think none of these do anything?  I already am sure a few do not, and I will confirm.  And no need for a generic list of env args?  Again, I can check the source and examples.

{E6E993DF-839D-44E1-8186-62C01AF1E662}

Stevenpc3: actually.. the ui-config that is created and mounted is passed in as an ENV and not a command line arg like the user-config is... so I assume that means that the ui-config does not even work if there are no ENV vars.

{6BC503F9-AE1F-488A-AA50-307CEF7D51F7}

yurishkuro: confirmed, none of these variables have any effect.

yurishkuro: UI config must be specified in the main yaml config (which can be overridden with `--set` as all yaml props)

Stevenpc3: well according to this doc https://www.jaegertracing.io/docs/2.14/deployment/frontend-ui/ it must be passed in as a pointer to json file via ```--query.ui-config``` commandline arg.  Which I guess is wrong now?  I guess the docs need updated.

The example here https://github.com/jaegertracing/jaeger/blob/main/cmd/jaeger/config.yaml#L39 DOES show it as a config key that points to a file though.

So I can clean those args up in the chart.

yurishkuro: the docs are wrong (stale), I'll update

yurishkuro: https://github.com/jaegertracing/documentation/pull/1057

Stevenpc3: this is clearly old (linked in the charts) https://www.jaegertracing.io/docs/1.76/deployment/cli/ but are there ANY cli flags (there is at least --config and --set) I guess I could try to run --help, but I don't have it deployed in a way that I can do that.  Do you have a link?  I can dig in the source also.

Stevenpc3: I added a commit that removes the elasticsearch and cassandra deps and cleaned up a lot.  It turns out there was a large number of broken/unused items.

I still need to do testing and perhaps simplify the spark / storage.xxx items.   And address some TODOs.

I might not get to this until Tuesday though.

charts/jaeger/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: 2.14.1
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
-version: 4.3.4
+version: 4.3.5
```
Stevenpc3: this should likely bump to 5.0.0 if the removal of things commit is kept.

charts/jaeger/Chart.yaml
```diff
@@ -27,14 +27,6 @@ maintainers:
...
 dependencies:
-  - name: cassandra
-    version: 0.15.3
-    repository: https://charts.helm.sh/incubator
-    condition: provisionDataStore.cassandra
-  - name: elasticsearch
-    version: 8.5.1
-    repository: https://helm.elastic.co
-    condition: provisionDataStore.elasticsearch
   - name: common
```
Stevenpc3, line 30: work to remove this.  It is part of image rendering.  pull in the common parts or rewrite.

charts/jaeger/Chart.yaml
yurishkuro: I would do 4.4.0

.github/workflows/lint-test.yaml
```diff
@@ -36,14 +36,3 @@ jobs:
       - uses: ./.github/actions/prepare-k8s
 
       - run: make test
-
-  test-with-elasticsearch:
```
yurishkuro, line 40: does it mean the ES specific configs are gone completely?

charts/jaeger/Chart.yaml
Stevenpc3: Seems fine since it's already labeled experimental.

charts/jaeger/values.yaml
yurishkuro: This is merged https://github.com/jaegertracing/jaeger/pull/7887

.github/workflows/lint-test.yaml
Stevenpc3: Yes for provisioning and connecting with Jaeger. There are still the maintenance jobs left because they are needed.

Testing here is not necessary as it is done through normal user configuration to external. Any testing for connectivity will be done in Jaeger and supported versions reflected in chart docs and Jaeger docs.

charts/jaeger/values.yaml
Stevenpc3: Dang, that was fast. 

That opens up the capability of not ripping out what I removed, but I do feel it's best to remove anyway. 

That MR still allows some more condensed configs. 

I guess think about for a few days then we can decide one way or the other.

charts/jaeger/Chart.yaml
yurishkuro:
```suggestion
version: 4.4.0
```

PR #738, opened 2026-02-11
[jaeger] Add extraVolumes/extraVolumeMounts to deployment

### LINKED ISSUE #737: [Bug]: Add extraVolumes/extraVolumeMounts support to jaeger deployment
### What happened?

 As a user deploying Jaeger v2 with a storage backend that requires local volumes (e.g. Badger), I want to be able to specify extra volumes and
  volume mounts on the jaeger deployment, so that I can mount PVCs, secrets, or other volumes without needing to fork the chart.

  The v4.2.0 rewrite simplified the chart to a stateless model targeting external storage (Elasticsearch, Cassandra) or in-memory for testing.
  However, Jaeger v2 still supports Badger as a storage backend via userconfig, and users can configure it to write to local directories — but
  there's no way to mount a volume at those paths since the deployment template only supports ConfigMap volumes for user-config and ui-config.

  extraObjects can't work around this either — it can create additional resources (PVCs, ConfigMaps, etc.) but can't inject volumes or
  volumeMounts into the chart's own Deployment template. The only workaround is setting jaeger.enabled: false and recreating the entire
  Deployment via extraObjects, which defeats the purpose of using the chart.

  Adding extraVolumes and extraVolumeMounts fields to the jaeger values section is a common Helm chart pattern (already used in this chart's
  spark, esIndexCleaner, esRollover, and esLookback jobs via extraSecretMounts/extraConfigmapMounts) and would cover this use case without
  re-adding dedicated Badger support.

### Steps to reproduce

1. Try and use badger with persistent storage

### Expected behavior

Have the ability to attach additional volumes as we could in 4.1.x

### Relevant log output

```shell

```

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

```shell

```

maynardflies:
## Summary
- Adds `extraVolumes` and `extraVolumeMounts` fields to the `jaeger` deployment values, restoring the ability to mount custom volumes (PVCs, secrets, configmaps, etc.) on the Jaeger pod
- The v4.2.0 chart rewrite simplified the deployment to only support ConfigMap volumes for user-config and ui-config, which broke storage backends like Badger that require persistent volumes
- Follows the same pattern already used by `spark`, `esIndexCleaner`, `esRollover`, and `esLookback` job templates

Fixes #737

## Test plan
- [x] `helm lint charts/jaeger` passes
- [x] `helm template` renders correctly with default values (no extra volumes)
- [x] `helm template` renders correctly with extra volumes/mounts set (PVC and mountPath appear in output)
- [x] CI values files render cleanly

### DIFF
--- a/charts/jaeger/Chart.lock
+++ b/charts/jaeger/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: common
   repository: https://charts.bitnami.com/bitnami
-  version: 2.34.0
-digest: sha256:63de10c1f68bedef391890484e1e29f2efd54e6dc7239ca60f0a572042efd4f3
-generated: "2026-02-03T13:47:26.844621Z"
+  version: 2.36.0
+digest: sha256:a9e48d71c1ab826e89a4115379550999c1787634d834acfe22122560393685b4
+generated: "2026-02-10T19:36:02.0023006+01:00"
--- a/charts/jaeger/Chart.yaml
+++ b/charts/jaeger/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 2.14.1
 description: A Jaeger Helm chart for Kubernetes
 name: jaeger
 type: application
-version: 4.4.7
+version: 4.4.8
 # Artifact Hub annotations
 # The jaeger image is whitelisted from security scanning because the reported
 # CVEs are in the upstream Alpine base image (OpenSSL libcrypto3/libssl3) and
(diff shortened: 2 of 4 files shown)


DevErenOzcan:
Hi! Thanks for working on this. I’m also looking forward to this change for my project. Do you have an estimate on when it might be ready? Let me know if there's anything I can help with!

